PR #16, opened 2017-09-21
Verify fixed-point decimal types

carver:
Added the fixed type described here: https://github.com/ethereum/wiki/wiki/Ethereum-Contract-ABI#types

### DIFF
--- a/eth_abi/utils/parsing.py
+++ b/eth_abi/utils/parsing.py
@@ -37,6 +37,17 @@ def process_strict_type(typ):
             raise ValueError("Integer size out of bounds")
         if int(sub) % 8 != 0:
             raise ValueError("Integer size must be multiple of 8")
+    # Check validity of fixed type
+    elif base == 'ufixed' or base == 'fixed':
+        if not re.match('^[0-9]+x[0-9]+$', sub):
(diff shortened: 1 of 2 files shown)


eth_abi/utils/parsing.py
```diff
@@ -37,6 +37,17 @@ def process_strict_type(typ):
             raise ValueError("Integer size out of bounds")
         if int(sub) % 8 != 0:
             raise ValueError("Integer size must be multiple of 8")
+    # Check validity of fixed type
+    elif base == 'ufixed' or base == 'fixed':
+        if not re.match('^[0-9]+x[0-9]+$', sub):
+            raise ValueError("Fixed type must have suffix of form <high>x<low>, eg. 128x128")
+        bits, minus_e = [int(x) for x in sub.split('x')]
```
pipermerriam, line 44: just a matter of preference when splitting something into exactly two pieces.

```python
bits, _, minus_e = sub.partition('x')
```

pipermerriam (approved):
Totally nitpicky change request if you don't mind.

eth_abi/utils/parsing.py
pipermerriam: oh, grumble, the casting to integer makes that more complex.  Ignore my suggestion.  Good to merge as-is.

PR #17, opened 2017-09-21
Automatically expand integer to 256, even if array

carver:


eth_abi/utils/parsing.py
```diff
@@ -1,18 +1,29 @@
...
     if typ == 'function':
         return process_strict_type('bytes24')
-    elif typ in ('int', 'uint'):
-        return process_strict_type(typ + '256')
-    else:
-        return process_strict_type(typ)
+
+    for base, length in DEFAULT_LENGTHS:
+        if typ.startswith(base):
+            type_with_length = _specify_length_if_absent(base, typ, length)
```
pipermerriam: This approach was really hard for me to read.  Can you refactor so that it looks something more like this?

```
if typ == 'function':
    return .....
elif is_missing_length(typ):
    default_length = get_default_length(typ)
    type_with_length = ....
    return process_strict_type(type_with_length)
else
    return process_strict_type(typ)
```

And now that I've written it like that, I'm thinking this could be even better.

```python
def process_type(raw_typ):
    typ = normalize_type(raw_typ)
    return process_strict_type(typ)


def normalize_type(raw_typ):
    if raw_typ == 'function':
        return 'bytes24'
    elif is_missing_length(raw_typ):
        default_length = get_default_length(raw_typ)
        type_with_length = ....
        return type_with_length
    else:
        return raw_typ
````

Mostly trying to make the functions easier to follow.  Do as you will.

pipermerriam (changes requested):
Looking to find a way to improve readability.

eth_abi/utils/parsing.py
carver: I was avoiding that because of all the shared logic in checking if the length is missing and then setting the default length. I think I found a good way to refactor it with a shared function call in both.

pipermerriam: I like it a lot more. 👍

pipermerriam (approved):

PR #18, opened 2017-09-26
process_type can handle bytes types again

carver:
`process_type()` used to support `raw_type` passed as `bytes`, but the new normalization did not support it.

This adds a test and readds support.


pipermerriam:
Release under `v0.4.3`

PR #20, opened 2017-10-26
showcase signed int decoder failure

carver:
The decoder implementation seems to expect `FF` padding instead of `00` padding on a negative integer. I don't know the details well enough here to fix this. (Not sure whether the test setup or implementation is wrong)

Apparently the failure has been there all along, but hypothesis just found it. I only added the new failing example in this PR to highlight it.

shrug

### DIFF
--- a/tests/decoding/test_single_decoders.py
+++ b/tests/decoding/test_single_decoders.py
@@ -4,6 +4,7 @@ import pytest
 
 import decimal
 from io import BytesIO
+import sys
 
 from hypothesis import (
     given,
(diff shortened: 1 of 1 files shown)


pipermerriam:
From: https://github.com/ethereum/wiki/wiki/Ethereum-Contract-ABI

screen shot 2017-10-26 at 4 43 03 pm

Looks like the test needs to be updated to detect this type of failure.  Unless you want to dig into this I'll take it from here.

carver:
Ok, got it. That should be enough to go on, I think.

carver:
*sigh* have I mentioned that I'm ready for Python2 to die? :expressionless:

PR #27, opened 2018-02-06
Carver/low precision decimal test

carver:
### What was wrong?

A version bump failed, that's weird:
https://travis-ci.org/ethereum/eth-abi/jobs/337782324 

Even weirder, a travis test on the same commit succeeded:
https://travis-ci.org/ethereum/eth-abi/jobs/337782341

Oh, it's a hypothesis test, so we had to get lucky on finding the example. Turns out the failure was related to building the expected value in the test, not in the project source.

### How was it fixed?

- add the failing example to make sure the case is always covered
- set precision higher when building expected value
- too many magic numbers: moved to centralized `decimal.Context` definition
- grep'd for all `Decimal` usages to put them in the `abi_decimal_context`

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/eth_abi/decoding.py
+++ b/eth_abi/decoding.py
@@ -13,6 +13,7 @@ from eth_abi.exceptions import (
     NonEmptyPaddingBytes,
 )
 from eth_abi.utils.numeric import (
+    abi_decimal_context,
     big_endian_to_int,
     quantize_value,
     ceil32,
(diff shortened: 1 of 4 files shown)


pipermerriam (approved):
Congratulations on unravelling this one.  Those hypothesis tests are somewhat insane to parse.

PR #37, opened 2018-03-15
Fix typos in exception messages

davesque:
### What was wrong?

Some exception messages contained typos.

### How was it fixed?

Fixed the typos.

#### Cute Animal Picture

Cute animal picture


davesque:
Rebuilt this branch after setting up commit signing.

PR #40, opened 2018-03-17
Fix issue with `test_decode_signed_real`

davesque:
### What was wrong?

The `test_decode_signed_real` property test fails when hypothesis gives
it a set leading data bit (encoding a negative number) and no set
padding bits or an unset leading data bit (a positive number) and fully
set padding bits.  The property test does not properly recognize such
cases as exception-raising.

Falsifying example:
```python
test_decode_signed_real(high_bit_size=16, low_bit_size=56, integer_bit_size=72, stream_bytes=b'\x00\x80\x00\x00\x00\x00\x00\x00\x00\x00', data_byte_size=10)
```

### How was it fixed?

Modified the utility function `is_invalid_padding_bytes` to more properly identify exception-raising cases.  Also renamed function to `is_valid_padding_bytes` since relevant specs tend to describe validity instead of invalidity.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/tests/decoding/test_single_decoders.py
+++ b/tests/decoding/test_single_decoders.py
@@ -59,15 +59,23 @@ def is_non_empty_non_null_byte_string(value):
     return value and big_endian_to_int(value) != 0
 
 
-def is_invalid_padding_bytes(value):
-    if not value:
-        return False
-    elif value.replace(b'\x00', b'') == b'':
-        return False
-    elif value.replace(b'\xff', b'') == b'':
-        return False
-    else:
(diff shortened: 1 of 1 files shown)


davesque:
This test happened to fail on my local dev machine with this falsifying example:

```python
test_decode_signed_real(high_bit_size=16, low_bit_size=56, integer_bit_size=72, stream_bytes=b'\x00\x80\x00\x00\x00\x00\x00\x00\x00\x00', data_byte_size=10)
```

PR #41, opened 2018-03-21
Add registry pattern to eth-abi

davesque:
### What was wrong?

Need to add registry pattern to eth-abi library.

### How was it fixed?

Not done yet.  Work in progress.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/eth_abi/registry.py
+++ b/eth_abi/registry.py
@@ -1,0 +1,383 @@
+import functools
+
+from . import (
+    decoding,
+    encoding,
+    exceptions,
+)
+
+from .base import BaseCoder
+
+from .utils import grammar
+
+
+class PredicateMapping:
+    """
+    Acts as a mapping from predicate functions to values.  Values are retrieved
+    when their corresponding predicate matches a given input.  Predicates can
+    also be labeled to facilitate removal from the mapping.
+    """
+    def __init__(self, name):
+        self._name = name
+        self._values = {}
+        self._labeled_predicates = {}
+
+    def add(self, predicate, value, label=None):
+        if predicate in self._values:
+            raise ValueError('Matcher {} already exists in {}'.format(
+                repr(predicate),
+                self._name,
+            ))
+
+        if label is not None:
+            if label in self._labeled_predicates:
+                raise ValueError(
+                    "Matcher {} with label '{}' already exists in {}".format(
+                        repr(predicate),
+                        label,
+                        self._name,
+                    ),
+                )
+
+            self._labeled_predicates[label] = predicate
+
+        self._values[predicate] = value
+
+    def find(self, type_str):
+        results = tuple(
+            (predicate, value) for predicate, value in self._values.items()
+            if predicate(type_str)
+        )
+
+        if len(results) == 0:
+            raise ValueError("No matching entries for '{}' in {}".format(
+                type_str,
+                self._name,
+            ))
+
+        predicates, values = tuple(zip(*results))
+
+        if len(results) > 1:
+            raise ValueError("Multiple matching entries for '{}' in {}: {}".format(
+                type_str,
+                self._name,
+                ', '.join(map(repr, predicates)),
+            ))
+
+        return values[0]
+
+    def remove_by_equality(self, predicate):
+        # Delete the predicate mapping to the previously stored value
+        try:
+            del self._values[predicate]
+        except KeyError:
+            raise KeyError('Matcher {} not found in {}'.format(
+                repr(predicate),
+                self._name,
+            ))
+
+        # Delete any label which refers to this predicate
+        try:
+            label = self._label_for_predicate(predicate)
+        except ValueError:
+            pass
+        else:
+            del self._labeled_predicates[label]
+
+    def _label_for_predicate(self, predicate):
+        # Both keys and values in `_labeled_predicates` are unique since the
+        # `add` method enforces this
+        for key, value in self._labeled_predicates.items():
+            if value is predicate:
+                return key
+
+        raise ValueError('Matcher {} not referred to by any label in {}'.format(
+            repr(predicate),
+            self._name,
+        ))
+
+    def remove_by_label(self, label):
+        try:
+            predicate = self._labeled_predicates[label]
+        except KeyError:
+            raise KeyError("Label '{}' not found in {}".format(label, self._name))
+
+        del self._labeled_predicates[label]
+        del self._values[predicate]
+
+    def remove(self, predicate_or_label):
+        if callable(predicate_or_label):
+            self.remove_by_equality(predicate_or_label)
+        elif isinstance(predicate_or_label, str):
+            self.remove_by_label(predicate_or_label)
+        else:
+            raise TypeError('Key to be removed must be callable or string: got {}'.format(
+                type(predicate_or_label),
+            ))
+
+
+class Predicate:
+    """
+    A class which represents a predicate function to be used for type matching
+    in ABIRegistry.
+    """
+    __slots__ = tuple()
+
+    def __call__(self, *args, **kwargs):  # pragma: no cover
+        raise NotImplementedError('Must implement `__call__`')
+
+    def __str__(self):  # pragma: no cover
+        raise NotImplementedError('Must implement `__str__`')
+
+    def __repr__(self):
+        return '<{} {}>'.format(type(self).__name__, self)
+
+    def __iter__(self):
+        for attr in self.__slots__:
+            yield getattr(self, attr)
+
+    def __hash__(self):
+        return hash(tuple(self))
+
+    def __eq__(self, other):
+        return (
+            type(self) is type(other) and
+            tuple(self) == tuple(other)
+        )
+
+
+class Equals(Predicate):
+    """
+    A predicate which matches any input equal to `value`.
+    """
+    __slots__ = ('value',)
+
+    def __init__(self, value):
+        self.value = value
+
+    def __call__(self, other):
+        return self.value == other
+
+    def __str__(self):
+        return '(== {})'.format(repr(self.value))
+
+
+class BaseEquals(Predicate):
+    """
+    A predicate which matches a basic type string with a base component equal to
+    `value` and no array component.  If `with_sub` is `True`, the type string
+    must have a sub component to match.  If `with_sub` is `False`, the type
+    string must *not* have a sub component to match.  If `with_sub` is None, the
+    type string's sub component is ignored.
+    """
+    __slots__ = ('base', 'with_sub')
+
+    def __init__(self, base, *, with_sub=None):
+        self.base = base
+        self.with_sub = with_sub
+
+    def __call__(self, type_str):
+        try:
+            abi_type = grammar.parse(type_str)
+        except exceptions.ParseError:
+            return False
+
+        if isinstance(abi_type, grammar.BasicType):
+            if abi_type.arrlist is not None:
+                return False
+
+            if self.with_sub is not None:
+                if self.with_sub and abi_type.sub is None:
+                    return False
+                if not self.with_sub and abi_type.sub is not None:
+                    return False
+
+            return abi_type.base == self.base
+
+        # We'd reach this point if `type_str` did not contain a basic type
+        # e.g. if it contained a tuple type
+        return False
+
+    def __str__(self):
+        return '(base == {}{})'.format(
+            repr(self.base),
+            '' if self.with_sub is None else (
+                ' and sub is not None' if self.with_sub else ' and sub is None'
+            ),
+        )
+
+
+def has_arrlist(type_str):
+    """
+    A predicate which matches a basic type string with an array dimension list.
+    """
+    try:
+        abi_type = grammar.parse(type_str)
+    except exceptions.ParseError:
+        return False
+
+    return (
+        isinstance(abi_type, grammar.BasicType)
+        and abi_type.arrlist is not None
+    )
+
+
+def _clear_encoder_cache(old_method):
+    def new_method(self, *args, **kwargs):
+        self.get_encoder.cache_clear()
+        return old_method(self, *args, **kwargs)
+
+    return new_method
+
+
+def _clear_decoder_cache(old_method):
+    def new_method(self, *args, **kwargs):
+        self.get_decoder.cache_clear()
+        return old_method(self, *args, **kwargs)
+
+    return new_method
+
+
+class ABIRegistry:
+    def __init__(self):
+        self._encoders = PredicateMapping('encoder registry')
+        self._decoders = PredicateMapping('decoder registry')
+
+    @staticmethod
+    def _register_coder(mapping, lookup, coder, label=None):
+        if callable(lookup):
+            mapping.add(lookup, coder, label)
+            return
+
+        if isinstance(lookup, str):
+            mapping.add(Equals(lookup), coder, lookup)
+            return
+
+        raise TypeError(
+            'Lookup must be a callable or a value of type `str`: got {}'.format(
+                repr(lookup),
+            )
+        )
+
+    @staticmethod
+    def _unregister_coder(mapping, lookup_or_label):
+        if callable(lookup_or_label):
+            mapping.remove_by_equality(lookup_or_label)
+            return
+
+        if isinstance(lookup_or_label, str):
+            mapping.remove_by_label(lookup_or_label)
+            return
+
+        raise TypeError(
+            'Lookup/label must be a callable or a value of type `str`: got {}'.format(
+                repr(lookup_or_label),
+            )
+        )
+
+    def _get_coder(self, mapping, type_str):
+        try:
+            coder = mapping.find(type_str)
+        except ValueError as e:
+            if 'No matching' in e.args[0]:
+                # If no matches found, attempt to parse in case lack of matches
+                # was due to unparsability
+                grammar.parse(type_str)
+
+            raise
+
+        if isinstance(coder, type) and issubclass(coder, BaseCoder):
+            return coder.from_type_str(type_str, self)
+
+        return coder
+
+    @_clear_encoder_cache
+    def register_encoder(self, lookup, encoder, label=None):
+        self._register_coder(self._encoders, lookup, encoder, label=label)
+
+    @_clear_encoder_cache
+    def unregister_encoder(self, lookup_or_label):
+        self._unregister_coder(self._encoders, lookup_or_label)
+
+    @_clear_decoder_cache
+    def register_decoder(self, lookup, decoder, label=None):
+        self._register_coder(self._decoders, lookup, decoder, label=label)
+
+    @_clear_decoder_cache
+    def unregister_decoder(self, lookup_or_label):
+        self._unregister_coder(self._decoders, lookup_or_label)
+
+    def register(self, lookup, encoder, decoder, label=None):
+        self.register_encoder(lookup, encoder, label=label)
+        self.register_decoder(lookup, decoder, label=label)
+
+    def unregister(self, lookup_or_label):
+        self.unregister_encoder(lookup_or_label)
+        self.unregister_decoder(lookup_or_label)
+
+    @functools.lru_cache(maxsize=None)
+    def get_encoder(self, type_str):
+        return self._get_coder(self._encoders, type_str)
+
+    @functools.lru_cache(maxsize=None)
+    def get_decoder(self, type_str):
+        return self._get_coder(self._decoders, type_str)
+
+
+registry = ABIRegistry()
+
+registry.register(
+    has_arrlist,
+    encoding.BaseArrayEncoder, decoding.BaseArrayDecoder,
+    label='has_arrlist',
+)
+registry.register(
+    BaseEquals('address'),
+    encoding.AddressEncoder, decoding.AddressDecoder,
+    label='address',
+)
+registry.register(
+    BaseEquals('bool'),
+    encoding.BooleanEncoder, decoding.BooleanDecoder,
+    label='bool',
+)
+registry.register(
+    BaseEquals('bytes', with_sub=True),
+    encoding.BytesEncoder, decoding.BytesDecoder,
+    label='bytes<M>',
+)
+registry.register(
+    BaseEquals('function'),
+    encoding.BytesEncoder, decoding.BytesDecoder,
+    label='function',
+)
+registry.register(
+    BaseEquals('bytes', with_sub=False),
+    encoding.ByteStringEncoder, decoding.ByteStringDecoder,
+    label='bytes',
+)
+registry.register(
+    BaseEquals('int'),
+    encoding.SignedIntegerEncoder, decoding.SignedIntegerDecoder,
+    label='int',
+)
+registry.register(
+    BaseEquals('string'),
+    encoding.TextStringEncoder, decoding.StringDecoder,
+    label='string',
+)
+registry.register(
+    BaseEquals('uint'),
+    encoding.UnsignedIntegerEncoder, decoding.UnsignedIntegerDecoder,
+    label='uint',
+)
+registry.register(
+    BaseEquals('ureal'),
+    encoding.UnsignedRealEncoder, decoding.UnsignedRealDecoder,
+    label='ureal',
+)
+registry.register(
+    BaseEquals('real'),
+    encoding.SignedRealEncoder, decoding.SignedRealDecoder,
+    label='real',
+)
--- a/eth_abi/utils/grammar.py
+++ b/eth_abi/utils/grammar.py
@@ -13,8 +13,12 @@ from eth_abi.exceptions import (
 grammar = parsimonious.Grammar(r"""
 type = tuple_type / basic_type
 
-tuple_type = "(" type next_type* ")"
+tuple_type = non_zero_tuple / zero_tuple
+
+non_zero_tuple = "(" type next_type* ")"
 next_type = "," type
+
+zero_tuple = "()"
 
 basic_type = base sub? arrlist?
 
@@ -39,7 +43,7 @@ class NodeVisitor(parsimonious.NodeVisitor):
     """
     grammar = grammar
 
-    def visit_tuple_type(self, node, visited_children):
+    def visit_non_zero_tuple(self, node, visited_children):
         # Ignore left and right parens
         _, first, rest, _ = visited_children
 
@@ -50,6 +54,9 @@ class NodeVisitor(parsimonious.NodeVisitor):
         _, abi_type = visited_children
 
         return abi_type
+
+    def visit_zero_tuple(self, node, visited_children):
+        return TupleType(tuple(), node=node)
 
     def visit_basic_type(self, node, visited_children):
         base, sub, arrlist = visited_children
--- a/eth_abi/utils/grammar.py
+++ b/eth_abi/utils/grammar.py
@@ -1,0 +1,344 @@
+import functools
+import re
+
+from parsimonious import expressions
+import parsimonious
+
+from eth_abi.exceptions import (
+    ABITypeError,
+    ParseError,
+)
+
+
+grammar = parsimonious.Grammar(r"""
+type = tuple_type / basic_type
+
+tuple_type = "(" type next_type* ")"
+next_type = "," type
+
+basic_type = base sub? arrlist?
+
+base = alphas
+
+sub = two_size / digits
+two_size = (digits "x" digits)
+
+arrlist = (const_arr / dynam_arr)+
+const_arr = "[" digits "]"
+dynam_arr = "[]"
+
+alphas = ~"[a-z]+"
+digits = ~"[1-9][0-9]*"
+""")
+
+
+class NodeVisitor(parsimonious.NodeVisitor):
+    """
+    Parsimonious node visitor which performs both parsing of type strings and
+    post-processing of parse trees.  Parsing operations are cached.
+    """
+    grammar = grammar
+
+    def visit_tuple_type(self, node, visited_children):
+        # Ignore left and right parens
+        _, first, rest, _ = visited_children
+
+        return TupleType((first,) + rest, node=node)
+
+    def visit_next_type(self, node, visited_children):
+        # Ignore comma
+        _, abi_type = visited_children
+
+        return abi_type
+
+    def visit_basic_type(self, node, visited_children):
+        base, sub, arrlist = visited_children
+
+        return BasicType(base, sub, arrlist, node=node)
+
+    def visit_two_size(self, node, visited_children):
+        # Ignore "x"
+        first, _, second = visited_children
+
+        return first, second
+
+    def visit_const_arr(self, node, visited_children):
+        # Ignore left and right brackets
+        _, int_value, _ = visited_children
+
+        return (int_value,)
+
+    def visit_dynam_arr(self, node, visited_children):
+        return tuple()
+
+    def visit_alphas(self, node, visited_children):
+        return node.text
+
+    def visit_digits(self, node, visited_children):
+        return int(node.text)
+
+    def generic_visit(self, node, visited_children):
+        if isinstance(node.expr, expressions.OneOf):
+            # Unwrap value chosen from alternatives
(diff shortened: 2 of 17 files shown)


davesque:
Code is currently untested.  Will probably add some tests before going any further.

eth_abi/registry.py
```diff
@@ -0,0 +1,146 @@
...
+        """
+        Registers an encoder for the given type lookup.
+        """
+        if callable(lookup):
+            # Is predicate lookup
+            self._encoder_predicates.add(lookup, encoder)
+            return
+
+        # Is string lookup
+        if lookup in self._encoder_registry:
```
pipermerriam: We've generally tried to write our conditionals to be explicit in the following manner.

```python
if callable(thing):
    # do something
elif isinstance(thing, str):
    # do something
else:
    raise TypeError("<good error message indicating type not supported, what type was given, what types are allowed")
```

It ensures we don't hit weird cases where someone uses `bytes` or an `int` or something and our registry ends up with weird values.

eth_abi/registry.py
```diff
@@ -0,0 +1,146 @@
...
+    def remove_matching(self, *args, **kwargs):
+        self._items = [i for i in self._items if not i.predicate(*args, **kwargs)]
+
+    def __contains__(self, item):
+        return item in tuple(i.predicate for i in self._items)
+
+
+class ABIRegistry:
+    def __init__(self):
+        self._encoder_registry = {}
```
pipermerriam: *naming nitpick*

Since the class is the registry, this seems more appropriately named `_encoders` since these are the registered encoders.

eth_abi/registry.py
```diff
@@ -0,0 +1,146 @@
...
+class NoEntryFound(ABIRegistryError):
+    pass
+
+
+PredicateItem = namedtuple('PredicateItem', ('predicate', 'value'))
+
+
+def ensure_changed(ExcClass, error_msg):
+    def decorator(old_method):
+        def new_method(self, *args, **kwargs):
```
pipermerriam: *naming convention nitpick*

We've got a somewhat consistent convention for decorators to use the names `fn`, `inner` and `outer`

```python
def simple_dec(fn):
    def inner(...):
         return fn(...)
    return inner

def args_dec(arg):
    def outer(fn):
         def inner(...):
             return fn(...)
        return inner
    return outer
```

I don't know if we have many in place that are method decorators, so no established convention for naming of the method fn.  I'd suggest something like `meth`, `method` or something else better I haven't though of.

eth_abi/registry.py
```diff
@@ -0,0 +1,146 @@
...
+
+    def match(self, *args, **kwargs):
+        return tuple(i.value for i in self._items if i.predicate(*args, **kwargs))
+
+    @ensure_changed(KeyError, 'Given predicate not found')
+    def remove_exact(self, predicate):
+        self._items = [i for i in self._items if i.predicate is not predicate]
+
+    @ensure_changed(KeyError, 'No matching items found')
+    def remove_matching(self, *args, **kwargs):
```
pipermerriam: This doesn't appear to be used.

eth_abi/registry.py
```diff
@@ -0,0 +1,146 @@
...
+        self._items = [i for i in self._items if not i.predicate(*args, **kwargs)]
+
+    def __contains__(self, item):
```
pipermerriam: Maybe I missed it but I don't see any `thing in predicate_list` code here (but I could have easily missed it).

pipermerriam (commented):
Generally good.  

I question the need for the `PredicateList` class.  Can we get away with only having the `_encoder_registry` and `_decoder_registry`?  It seems like we should be able to key on the `lookup`.

- custom functions/callables can just be identified by `id`
- string lookups can be either pre-processed into a callable which is unique based on the original string **or** we can just store the string and deal with the matching at some later stage when we try to lookup an encoder/decode.

Thoughts?

eth_abi/registry.py
davesque: Yep, things aren't done yet so I just have that in there in case I want to unregister callbacks in a different way.

eth_abi/registry.py
davesque: Sure, I can use that convention.  The naming convention I've been using is geared toward being as descriptive as possible but still reasonably concise which is hopefully helpful to people who are less strong with decorators.  I've seen names like `outer`, `inner` but I generally don't find them to be very meaningful.  But I'm happy to go with that if it's already an established pattern.

eth_abi/registry.py
davesque: You're right.  I need to take that out.  I had previously been doing explicit checks for membership somewhere but then switched to catching key errors instead.

eth_abi/registry.py
davesque: Good suggestion.  The naming there has to do with the fact that I originally had a single dict just called `_registry`.  I'll change it.

eth_abi/registry.py
davesque: Sounds good.  Just FYI, part of my deciding to do things that way had to do with similar duck-typing I saw here:
https://github.com/ethereum/eth-abi/blob/master/eth_abi/abi.py#L26-L29

Should we consider being more explicit in that case as well?

davesque:
I probably should have mentioned that I had a chat with Jason this morning and we had the thought that it seemed nicer to just be able to register a simple string representation of a type (e.g. `"uint256"`) instead of a tuple of parsed components (e.g. `(base, sub, arrlist)`).  Regardless, it seemed like it would be nice to have a separate dictionary in which mappings from simple hashable lookups (whether strings or tuples of parsed parts) to encoders and decoders could be quickly traversed (avoiding the need to iterate through and execute a list of functions).

However, it does seem that a dictionary would work just fine to store our predicate functions since functions are hashable (as you mention) and since we require that only one applies to the type which is being looked up (so order doesn't matter).  Also, we can convert simple string or tuple lookups to their equivalent predicate functions transparently (which you also mentioned).  In that case, I'd like to memoize the method on the registry class which does the lookup since it seems slow to iterate through a list of functions whenever we want to lookup an encoder or decoder.  Maybe we can clear the memo cache whenever someone registers or unregisters an encoder or decoder?

eth_abi/registry.py
pipermerriam: Yes, that would be an improvement.  Adding type hints here (and mypy typing would also help but that should probably be done independently).

eth_abi/registry.py
pipermerriam: I can get onboard with that.  +1 (no need to go update the other cases, we can just migrate our convention over time when that code gets touched).

davesque:
Did a bit of a rewrite after your feedback.  Lots of things are obviously not done and everything's pretty rough.

The one thing I did which I anticipate you may not like is to have a separate helper class like last time called `PredicateMapping`.  The idea with that class is that it's supposed to encapsulate a mapping to values by matching predicates and also handles some stuff I came up with for working with labels (though credit for the idea of having labels goes to @carver).  I was skeptical at first, but after thinking about it, it does seem easier from a user's point of view to juggle labels for predicate functions instead of hanging on to references for the functions themselves in case someone wants to unregister one.

davesque:
I finally came around on the labelling idea @carver. :)

eth_abi/registry.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def register_encoder(self, lookup, encoder, label=None):
+        """
+        Registers an encoder under the given type lookup and optional label.
+        """
+        if callable(lookup):
+            self._encoders.add(lookup, encoder, label)
+            return
+
+        if isinstance(lookup, str):
```
carver: This is a little more awkward with the predicate approach, but probably still worth doing: 

We talked about checking the `str` against existing `str` predicates at register-time. Unfortunately, I can't think of a better way to check the predicates than at encode time.

eth_abi/registry.py
```diff
@@ -0,0 +1,94 @@
...
+class PredicateMapping:
+    def __init__(self):
+        self._values = {}
+        self._labeled_predicates = {}
+
+    def add(self, predicate, value, label=None):
+        if label is not None:
+            self._labeled_predicates[label] = predicate
+
+        self._values[predicate] = value
```
carver: I think this should raise if the label or predicate already exist.

carver (commented):
Direction seems good to me. Just a couple safety checks to consider.

eth_abi/registry.py
davesque: Yep, thanks for the heads up.

eth_abi/registry.py
davesque: Need some clarification on this.  Are you wondering how we'll detect conflicts at register time if everything is a predicate?

carver: Actually, I missed that this was adding the string `lookup` as a label. So resolving the other comment will address this issue.

carver: I was just wanting to make sure that registering encoders twice with the same string would fail.

eth_abi/registry.py
```diff
@@ -0,0 +1,100 @@
...
+    def register_encoder(self, lookup, encoder, label=None):
+        """
+        Registers an encoder under the given type lookup and optional label.
+        """
+        if callable(lookup):
+            self._encoders.add(lookup, encoder, label)
+            return
+
+        if isinstance(lookup, str):
+            def predicate(base, sub, arrlist):
```
pipermerriam: Extracting this to the module level and currying it should make this a bit cleaner.

```python
from cytoolz import curry

@curry
def static_predicate(lookup, base, sub, arrlist):
    return collapse_type(base, sub, arrlist) == lookup

... # then using it later.
self.encoders.add(static_predicate(lookup), encoder, lookup)
```

davesque: I guess that means we'd need to add cytoolz under `install_requires` in `setup.py`.  Are there any problems that might come up as a result of both eth-abi and eth-utils depending on cytoolz (e.g. different version specs for cytoolz package might become out of sync, etc.)?  I guess I've worked on other projects where this might have come up but never thought of this before.

pipermerriam: `cytoolz` is pretty standard across our codebases as well as being quite stable.  Maybe worth doing a quick check that the version ranges intersect with whatever is specified in `eth-utils`

eth_abi/registry.py
```diff
@@ -0,0 +1,126 @@
...
+        Unregisters the encoder associated with the given type lookup or label.
+        """
+        if callable(lookup_or_label):
+            try:
+                self._encoders.remove_by_reference(lookup_or_label)
+            except KeyError:
+                raise NoEntryFound('No entry found for given predicate')
+            return
+
+        if isinstance(lookup_or_label, str):
```
davesque: @pipermerriam Should I maybe be using `eth_utils.is_text` here?

pipermerriam: I don't think it's needed.  Now that we're in python3 and differentiating between text/bytes doesn't require a check for what python version we're in, I think we can move away from `is_text` and `is_bytes` and potentially even `is_string`.  For lower level libs like this one, reducing the call-stack overhead should yield some basic albeit minor performance gains.

eth_abi/registry.py
```diff
@@ -0,0 +1,131 @@
+from cytoolz import curry
+
+from .utils.parsing import collapse_type
+
+
+class ABIRegistryError(Exception):
```
pipermerriam: We've typically located these in `lib_name.exceptions` by convention.

eth_abi/registry.py
```diff
@@ -0,0 +1,131 @@
...
+
+        if label is not None:
+            if label in self._labeled_predicates:
+                raise ValueError('Predicate with label "{}" already exists'.format(label))
+
+            self._labeled_predicates[label] = predicate
+
+        self._values[predicate] = value
+
+    def find(self, *args, **kwargs):
```
pipermerriam: Is the `*args, **kwargs` here because you aren't yet sure what the arguments will be, or to make this *generic*?  I think I'm ok with us making this implementation specific to whatever we *know* the arguments to our `predicate` functions are going to be so that we can eliminate one axis of validation/errors that this is able to produce by forcing things to fail at the `find(..)` invocation rather than somewhere deeper in the `predicate(...)` invocation.

eth_abi/registry.py
```diff
@@ -0,0 +1,131 @@
...
+        results = tuple(
+            value for predicate, value in self._values.items()
+            if predicate(*args, **kwargs)
```
pipermerriam: *nitpick*

I think that this tuple comprehension is equivalent to the [`toolz.keyfilter`](http://toolz.readthedocs.io/en/latest/api.html#toolz.dicttoolz.keyfilter).  Always questionable whether using such functions is an improvement as they are often a readability trade off.  Merely pointing it out so you know the tool exists.

eth_abi/registry.py
```diff
@@ -0,0 +1,131 @@
...
+            label = self._label_for_predicate(predicate)
+        except ValueError:
+            pass
+        else:
+            del self._labeled_predicates[label]
+
+    def _label_for_predicate(self, predicate):
+        for key, value in self._labeled_predicates.items():
+            if value is predicate:
+                return key
```
pipermerriam: Does this data structure contain an *implicit* guarantee that this loop either matches 1 or 0 values and cannot match multiple predicates?  If so, it might be worth adding a comment here to ensure future readers of this code know that case is not something that was *overlooked*.

tests/registry/test_predicate_mapping.py
```diff
@@ -0,0 +1,114 @@
...
+    PredicateMapping,
+)
+
+
+@pytest.fixture
+def mapping():
+    return PredicateMapping()
+
+
+def gt(x, y):
```
pipermerriam: Are you familiar with the [`operator`](https://docs.python.org/3/library/operator.html#operator.gt) module?

tests/registry/test_predicate_mapping.py
```diff
@@ -0,0 +1,114 @@
...
+    return x <= y
+
+
+def test_can_add_predicates_and_find_corresponding_values(mapping):
+    mapping.add(gt, 'greater than')
+    mapping.add(lte, 'less than or equal')
+
+    assert mapping.find(2, 1) == 'greater than'
+    assert mapping.find(2, 2) == 'less than or equal'
+    assert mapping.find(2, 3) == 'less than or equal'
```
pipermerriam: This isn't a big deal at all since these are so simple, but traditionally these would be done with the `pytest.mark.parametrize` decorator.  Just making sure you're aware of it's existence/usage.

tests/registry/test_predicate_mapping.py
```diff
@@ -0,0 +1,114 @@
...
+    e.match('Predicate already exists')
+
+
+def test_cannot_add_two_predicates_with_same_label(mapping):
+    mapping.add(gt, 'foo', 'my_label')
+
+    with pytest.raises(ValueError) as e:
+        mapping.add(lte, 'bar', 'my_label')
+
+    e.match('Predicate with label')
```
pipermerriam: Are you familiar with the `match` kwarg that `pytest.raises` accepts?

https://docs.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions

It should allow this to be written more succinctly as:

```python
with pytest.raises(ValueError, match="Predicate with label * already exists"):
    mapping.add(lte, 'bar', 'my_label')
```

eth_abi/registry.py
davesque: It was to make it generic but we can make it specific to its intended use.  I'll change it.

eth_abi/registry.py
davesque: I did notice that and planned to move the exceptions later.  Was just keeping everything in one file while I work on it.  But I'll go ahead and move it there now.

eth_abi/registry.py
davesque: It does give that guarantee since both keys and values in that dict must be unique (or an exception is raised).  I'll add a comment to that effect.

tests/registry/test_predicate_mapping.py
davesque: Yes I am.  I can use it here if you think that makes things more readable.

tests/registry/test_predicate_mapping.py
davesque: I did notice you guys using that decorator and thought I might use it in a case where there are many repetitions of some assertion.  I can use it there it you'd like.

tests/registry/test_predicate_mapping.py
davesque: I wasn't familiar with that.  I haven't used pytest extensively so I'm happy to hear about any cool bells and whistles.

tests/registry/test_predicate_mapping.py
davesque: Actually that stuff is going to get taken out since I'm assuming only one argument for predicates.  I had a discussion with @carver and he'd like to use type strings as the canonical form for types.  So that means predicates would only ever receive one string as an argument.  They can parse that string as much as they like in order to figure out if they're interested in it.  We're planning on memoizing the type lookup so the added performance cost of the repeated parsing shouldn't be a problem.  I'm fine with going in that direction.  Thoughts?

tests/registry/test_predicate_mapping.py
davesque: Actually I'm remembering now that the reason I did that was I thought it would be nice to test if the `find` method works if more than one predicate is present in the mapping.

eth_abi/registry.py
```diff
@@ -0,0 +1,127 @@
...
+            raise ValueError('Multiple matching values found')
+
+        return results[0]
+
+    def remove_by_reference(self, predicate):
+        # Delete the predicate mapping to the previously stored value
+        try:
+            del self._values[predicate]
+        except KeyError:
+            raise KeyError('Predicate not found')
```
pipermerriam: Seems like including the `str` representation of the `predicate` in the error here would be beneficial under the hope that our predicate functions are at least sometimes named in helpful ways.

eth_abi/registry.py
```diff
@@ -0,0 +1,127 @@
...
+        del self._labeled_predicates[label]
+        del self._values[predicate]
+
+    def remove(self, predicate_or_label):
+        if callable(predicate_or_label):
+            self.remove_by_reference(predicate_or_label)
+        elif isinstance(predicate_or_label, str):
+            self.remove_by_label(predicate_or_label)
+        else:
+            raise TypeError('Key to be removed must be a callable or a string')
```
pipermerriam: This is another place where including the type that was received in the error message makes debugging this failure path easier later.  Something like `"...: Got {0}".format(type(predicate_or_label))`

pipermerriam (approved):

eth_abi/registry.py
davesque: When you say string representation, do you just mean `str(func_or_lambda)`, so something like this: `<function foo at 0x110b416a8>` or `<function <lambda> at 0x110b38bf8>`?

davesque: Just want to make sure you're not talking about using something like the `inspect` module. :)

davesque: Just ignore the comments I made in reply to this.  I hadn't read your comment thoroughly enough.

pipermerriam: You are correct that I mean `str(func_or_lambda)`.  I think this would be well complimented by changing the `equals` function as follows.

```python
def equals(value):
    def inner(other):
        return other == value
    inner.__name__ = "is_equal_to_{0}".format(value)
    return inner
```

This way our auto-generated functions will have helpful string representations.

One thing that isn't ideal about this is that `equals('arst') != equals('arst')` since the generated functions won't be equivalent.  I think we can fix this with some class shenanigans.

```python
class Equals:
    def __init__(self, value):
        self.value = value
    def __call__(self, other):
        return self.value == other
    def __eq__(self, other):
        return isinstance(other, Equals) and self.value == other.value
```

There are probably a few more dunder methods that need to be implemented like `__hash__`.

davesque: It is possible that we can forego all of that by supplying labels for the default predicates?  My thinking was that, if people wanted to unregister default behavior, they would do so via one of the informatively chosen labels that we include.  I was also thinking of having a method on the registry that will provide a nice summary of what predicates are registered and any labels they might have.

If not, I think the idea of a callable `Equals` class is kinda cool anyway and I'd be happy to go with it.  It just seems like we could potentially avoid some of the edge cases that might come out of that approach by deferring to the labeling functionality.

Thoughts?

pipermerriam: >  I was also thinking of having a method on the registry that will provide a nice summary of what predicates are registered and any labels they might have.

Implementing this with `__str__` seems appropriate.

davesque: Went ahead and just took your suggestion here of having the `Equals` class.

eth_abi/utils/grammar.py
```diff
@@ -0,0 +1,270 @@
...
+                comp_str=node.text,
+                col=node.start + 1,
+                type_str=node.full_text,
+                error_msg=error_msg,
+            ),
+        )
+
+
+class TupleType(ABIType):
+    __slots__ = ('components',)
```
pipermerriam: omg.  How did I not know about this!

http://book.pythontips.com/en/latest/__slots__magic.html

Looks like a 40-50% reduction in memory footprint!

pipermerriam (commented):
Looks like this is coming along.  It wasn't entirely clear to me reading the code what all of the various classes are for.  Wondering if you could take a few minutes to write up some basic docstrings where you think it makes sense.

eth_abi/utils/grammar.py
davesque: Yeah it seems like `__slots__` is ideal for small "immutable" container classes that only have a few attributes and many instances.  Fast constant time attribute lookups and space savings ftw.

davesque: Although Python dicts in 3.6 share keys and are compact.  So I think it's less and less of a problem to just use normal classes with `__dict__` attrs.  Using `__slots__` is still a neat trick to have in your bag though.

pipermerriam: Also, it's looking like our `mypy` based type checking is going to be incompatible with `__slots__`.  See https://github.com/ethereum/py-evm/issues/520

davesque: Is that still the consensus?  It seemed that @carver had found a way to make it work.

pipermerriam: Fine to leave in for now until we implement type hinting at which point we can see if we can use @carver 's solution.

eth_abi/registry.py
```diff
@@ -0,0 +1,254 @@
...
+        if isinstance(lookup_or_label, str):
+            try:
+                mapping.remove_by_label(lookup_or_label)
+            except KeyError:
+                raise exceptions.NoEntryFound('No entry found for given label')
+            return
+
+        raise TypeError('Lookup/label must be a callable or a value of type `str`')
+
+    @classmethod
```
pipermerriam: looks like this could be `staticmethod`

eth_abi/registry.py
```diff
@@ -0,0 +1,254 @@
...
+            mapping.add(lookup, coder, label)
+            return
+
+        if isinstance(lookup, str):
+            mapping.add(Equals(lookup), coder, lookup)
+            return
+
+        raise TypeError('Lookup must be a callable or a value of type `str`')
+
+    @classmethod
```
pipermerriam: looks like this could be `staticmethod`

eth_abi/registry.py
```diff
@@ -0,0 +1,254 @@
...
+            ),
+        )
+
+
+class ABIRegistry:
+    def __init__(self):
+        self._encoders = PredicateMapping()
+        self._decoders = PredicateMapping()
+
+    @classmethod
```
pipermerriam: looks like this could be `staticmethod`

eth_abi/utils/grammar.py
```diff
@@ -0,0 +1,325 @@
...
+        if isinstance(node.expr, expressions.Optional):
+            # Unwrap optional value or return `None`
+            if len(visited_children) != 0:
+                return visited_children[0]
+
+            return None
+
+        return tuple(visited_children)
+
+    @functools.lru_cache(maxsize=None)
```
pipermerriam, line 101: Any reason for the `None`?  Seems like it should be fine to put an arbitrary upper bound on this (1024?) so that we **know** it can't grow indefinitely.

pipermerriam:
It *might* be good to elevate all of those `staticmethod` functions to just be top level function definitions to reduce the size of that class as well as the code complexity for understanding what the class is doing.

eth_abi/utils/grammar.py
davesque: I just figured that, in most systems using eth-abi, there would probably only be on the order of a few hundred type strings at most which would ever be parsed.  Also, with `maxsize=None`, `functools.lru_cache` foregoes a lot of machinery that it would normally use to properly evict things from the cache.  Since I figured the cache would probably never grow too large, I thought it would be better to opt for the speedier version which doesn't limit size.

davesque:
I suppose I would move them to the top level but it does seem like those methods are only really relevant to the way that the `ABIRegistry` class works internally.  So I feel like making that change might be confusing for some readers and may incorrectly imply that the methods have some more general usage.  Thoughts?

davesque:
Squashed some commits and rebased onto master.

pipermerriam:
> So I feel like making that change might be confusing for some readers and may incorrectly imply that the methods have some more general usage. Thoughts?

I don't feel strongly.  It's more of a stylistic choice.  Make whatever call you feel is best.

```python
def _do_thing(values):
    ...


class ThingDoer:
    @staticmethod
    def _do_thing(values):
        ...

    # using staticmethod
    def make_it_so(self, values):
        ...
        done_thing = self._do_thing(vaues)
        ...
        return ...

    # without staticmethod
    def make_it_so(self, values):
        ...
        done_thing = _do_thing(values)
        ...
        return ...
```

I like the module level function because I know that `_do_thing` doesn't have access to anything but the `values` it was passed.  Conversely, when I see `self._do_thing` I don't know the limits of the method call until I've looked to see that it's a `staticmethod`.

Because of this, I've *tended* to only use `staticmethod` as a way to expose public convenience APIs.  For internal purposes, I've tended to use top level functions *or* something relegated to `eth_abi.utils` if it warrants having tests written for it.

davesque:
@pipermerriam @carver 

I've had some thoughts about the `from_type_str` API and the proposed `FromTypeStr` mixin class which will be used to detect it in the registry (see [here](https://github.com/davesque/eth-abi/blob/cc8e5bdf0c5fc0e1b7a5cf51ba90bbb69ef04789/eth_abi/base.py)).  Earlier, a funny thing happened.  I added `FromTypeStr` to the inheritance list of `BytesEncoder` and implemented `_from_type_str`<sup>[1](#footnote)</sup>.  However, I then went and implemented `_from_type_str` on `BytesDecoder` but forgot to add `FromTypeStr` in its inheritance list.  I then got a rather unhelpful error message when the registry didn't find `FromTypeStr` in the inheritance list of `BytesDecoder` and failed to calls its implementation of `_from_type_str`.

It seems a bit silly to check that something is a subclass of `FromTypeStr` if all we're really interested in is whether or not it implements `_from_type_str`.  It seems pretty likely that someone might implement `_from_type_str` (because that's really the important thing to do) but forget to go through the motions of adding `FromTypeStr` to an inheritance list.

Would it make sense to forego having a `FromTypeStr` mixin altogether and just check if an object has a `_from_type_str` attribute?  I guess one nice thing about have a `FromTypeStr` mixin is that we can just point to it and say, "There, implement everything in that class to make this work."  But then it seems like a necessary but error prone formalism to check that an object is a subclass of `FromTypeStr`.

Thoughts?

<a name="footnote">1</a>: I renamed `from_type_str` to `_from_type_str` since Piper and I had a conversation earlier today about not yet exposing the registry as part of the public API.

pipermerriam:
Is it out of the question to just require that API be in place on all encoders/decoders?  We could even provide a default implementation in the base class which just naively instantiates and returns an instance of the class.

tests/property/test_predicates.py
```diff
@@ -46,6 +47,7 @@ def test_equals_has_expected_repr():
 
 @settings(max_examples=200)
 @given(type_strs)
+@example('int0')
```
pipermerriam: Is `int0` a valid type ?  in fact is anything with a sub of `0` valid?

davesque: No, it's not.  But it's parsable.  I'm pretty sure it could be parsed by the regexes we were using as well.  I guess I could update the grammar so that "0" is not a parsable sub component.  But it seemed like we were already drawing a pretty clear line between parsability and validity.

davesque:
I'm fine with that idea.  I guess that would mean there'd be some default implementation on the base encoder and decoder classes instead of on a mixin class, correct?

pipermerriam:
> there'd be some default implementation on the base encoder and decoder classes instead of on a mixin class, correct?

Yes, but now that we're talking about it I think a default implementation might be a foot gun, since *most* encoders are likely to need to set some values based on bitsizes, or whatever.

davesque:
Actually, default implementation was not what I meant.  That was a poor choice of words.  What I was really thinking was a stub implementation that raises when called.

davesque:
Left to do:
* Add support for array encoders in registry
* Add thorough testing of default type registrations
* Swap in registry code in place of old type parsing and routing
* Profit 💰💰

davesque:
Okay, everything's swapped in and all existing tests are passing.  But I'd still like add more tests for the code that I've written most recently.  There are some basic tests in place but still not enough.

davesque:
@carver @pipermerriam 

Rebuilt branch since it was getting a bit untidy.  I think things are mostly ready at this point.  All the new modules have good coverage.  The public API functions now use the registry internally.  As a reminder to @carver , with this latest version of the branch I was aiming to get things working well enough so that we can move on to other things and unblock some of the casper people (per a convo I had with Piper).

I did a grep on all of the 25 ethereum foundation python projects that I have access to.  The only ones which appear use `eth-abi` are `eth-tester` and `web3.py`.  The `eth-tester` project only uses the functions which are imported directly from `eth-abi` (i.e. `from eth_abi import ...`).  The `web3.py` project uses those same functions and also imports `process_type` and `collapse_type` from `eth_abi.abi`.  I made sure that there are backwards compatible versions of those functions which are still importable from that location.

Things left to do in the long term:
* Documentation
* Maybe another run of polish to get things in shape for public use
* Possibly move validation entirely into coder classes.  It's seeming more and more like coder classes should be responsible for validating type strings.  This would happen after the registry had routed a type string to a specific coder class.  That would mean coders are not only responsible for coding information to and from a type; they're also responsible for determining if meta information about a type (contained in a type string) makes sense.

pipermerriam (approved):

pipermerriam:
Looks good to merge.  I'll cut a beta release with this once merged and we can do some live testing against web3 to make sure we're not going to break things.

carver (approved):

carver:
Frankly, my eyes glazed over during the hypothesis strategy setup. Everything else looks good :+1:

carver:
> The web3.py project uses those same functions and also imports process_type and collapse_type from eth_abi.abi. I made sure that there are backwards compatible versions of those functions which are still importable from that location.

That's great. Also, I'm happy to migrate web3 to the newer parsed format at some point. It's not at the top of my list right now, but if it helps remove some cruft from eth-abi, then I'm all for it.

pipermerriam:
@carver regarding migrating to the new `eth-abi` stuff.  I think that implies allowing users to swap out the `encode_abi` and `decode_abi` functions that are used by web3.  This might be best done by allowing users to set which *registry* they are using.

carver:
> regarding migrating to the new eth-abi stuff. I think that implies allowing users to swap out the encode_abi and decode_abi functions that are used by web3. This might be best done by allowing users to set which registry they are using.

@pipermerriam that's cool too, and seems like a good way to do it. What I meant to say was to stop using `process_type()` and `collapse_type()` and instead use `eth_abi.grammar.parse()`

PR #42, opened 2018-03-23
Experimental (crazy) approach to allowing extensible parsing

davesque:
Just some ideas for how we could implement extensible parsing.

### What was wrong?

Was trying to figure out interesting methods of extensibility for the eth-abi library.

### How was it fixed?

Added a parser defined with `parsimonious` library and included the beginnings of some functionality for extending that parser.

It might be possible to allow users to define their own parsing rules for types which can't be parsed with our default grammar.  They could then create a subclass of `ABIType` to handle the processing of parse results for those rules.  The `ABIType.from_node` static method iterates through all subclasses of `ABIType` (the ordering, I believe, goes from most recently defined to least recently defined) to find one which can handle the parsing results for a certain rule (identified by the `expr_name` property on a parsimonious parse node).  `ABIType` subclasses indicate their associated grammar rule via the `expr_name` class property.  They must implement `from_node` as a class method in order to produce an instance of themselves according to the results of their associated parse rule.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/eth_abi/grammar.py
+++ b/eth_abi/grammar.py
@@ -1,0 +1,117 @@
+import parsimonious
+
+
+user_types = r"""
+user_types = ""+
+"""[1:-1]
+
+
+parser = parsimonious.Grammar(r"""
+type = tuple_type / basic_type / user_types
+
+{user_types}
+
+tuple_type = "(" type next_type* ")"
+next_type = "," type
+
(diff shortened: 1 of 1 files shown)


davesque:
@pipermerriam says there's not really a need for this currently.  Just wanted to add it here for historical reference.

PR #43, opened 2018-04-04
Fix abi_type_strategies.sized_list_strats

davesque:
### What was wrong?

The values of `type_str` and `type_strat` are not determined until the lambdas are evaluated. 
Therefore, every resulting lambda references the values in the last tuple in `all_basic_raw_strats` which happen to be for address sampling.

This causes the resulting strategy to only generate *address* lists instead of a list for any possible type in `all_basic_raw_strats`.

Also, the shared value key should probably be more unique.

### How was it fixed?

Said values are now explicitly captured in lambdas.  Also changed shared value key.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/tests/property/abi_type_strategies.py
+++ b/tests/property/abi_type_strategies.py
@@ -84,12 +84,12 @@ sized_list_strats = [
     st.tuples(
         st.shared(
             st.integers(min_value=MIN_LIST_SIZE, max_value=MAX_LIST_SIZE),
-            key="n",
-        ).map(lambda n: type_str + "[{0}]".format(n)),
+            key="sized_list_strats",
+        ).map(lambda n, type_str=type_str: type_str + "[{0}]".format(n)),
         st.shared(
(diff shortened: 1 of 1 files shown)


carver:
Interesting, I couldn't have told you what this would do without trying it:

```py
> [m() for m in [lambda: v for v in range(3)]]
[2, 2, 2]
```

So the idea is that the lambda closes around `v` and the for loop mutates v, so all three lambdas here return the final value, right?

But this works, because it would invoke before `v` changes:

```py
> [(lambda: v)() for v in range(3)]
[0, 1, 2]
```

Good catch!

carver (approved):

davesque:
@carver Yep.  Actually, my choice of words was poor and the way you're talking about it is more correct.  The problem is not that the variables are not closed over.  Quite the opposite.  Adding a default value for a keyword argument forces Python to evaluate the variable when the lambda is defined.

Issue #39, opened 2018-03-17
Add support for Vyper decimal10 ABI type


davesque:
### What was wrong?

Need to add support for Vyper `decimal10` type.  See here: https://github.com/ethereum/web3.py/issues/700

### How can it be fixed?

Need to add support for it.

davesque:
https://github.com/ethereum/web3.py/issues/700

davesque:
For the record, PR #48 was empty and accidentally marked as merged by github.

davesque:
PR #49 added support for solidity fixed types which could be useful in implementing support for vyper decimal types.  There's a fair bit of discussion surrounding that here: https://github.com/ethereum/vyper/issues/566

We need to figure out if we're going to move forward with support for decimal types based on fixed types or if we're going to wait for some kind of decision to be made about ditching decimal types entirely in favor of fixed types.

davesque:
At present, I believe the plan is for Vyper to use the ABI `fixed<M>x<N>` type at the appropriate precision.  Going to close this issue.

PR #48, opened 2018-04-09
Add support for Vyper decimal10 ABI type

davesque:
*Note: This branch is rooted at the current `registry` branch head.*

### What was wrong?

Must add support for Vyper `decimal10` type.  Issue ticket here:
https://github.com/ethereum/eth-abi/issues/39

Cute animal picture

### DIFF
--- a/eth_abi/abi.py
+++ b/eth_abi/abi.py
@@ -7,14 +7,10 @@ from eth_utils import (
     is_bytes,
 )
 
-from eth_abi.decoding import (
-    get_single_decoder,
-    get_multi_decoder,
-)
-from eth_abi.encoding import (
-    get_single_encoder,
-    get_multi_encoder,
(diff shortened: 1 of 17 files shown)


pipermerriam:
I must be missing something.  I don't see any code here for `decimal` type support?

carver:
Yeah, I think github auto-closed this one because it had the same tip as the registry PR.

davesque:
Yep sorry about that.  I meant to mention this in today's standup.  I had opened an empty PR which I guess github let me do since it was based on the head of the unmerged `registry` branch.  But then, when I merged in the registry, github automatically closed the vyper pull request.  I guess there's not really any way to open an empty PR.

pipermerriam:
It's ok, I was just confused.  Makes sense.

PR #49, opened 2018-04-11
Add Solidity fixed types

davesque:
### What was wrong?

Must add support for Vyper `decimal10` type.  Issue ticket here: https://github.com/ethereum/eth-abi/issues/39

#### Cute Animal Picture

Cute animal picture


davesque:
It seemed like there are so many crossovers between decimal and fixed types that I just went ahead and implemented the more general fixed type.  I should be able to create decimal coders by subclassing fixed coders and just explicitly setting `value_bit_size = 168` and `frac_places = 10`.  See [here](https://github.com/ethereum/vyper/issues/566) for a justification.  I'm going to double check the math in that Vyper issue ticket just to be dilligent and make sure those are actually the correct settings.

Left to do:
* Write property tests for the fixed decoders
* Add strategies for end-to-end testing of encoded fixed values
* Add decimal coders as subclasses of fixed coders

I don't think we'll need to test decimal coders very much since the coder subclasses shouldn't need to override many of the methods on their parent classes.

pipermerriam:
👍 largely looks good.  Lemme know when this is ready for merge/review.

carver:
> This is a problem though since it is 6 bits larger than the underlying Viper type.

I'm curious to hear how that affects the plan.

IMO the decimal type doesn't add much value beyond the fixedMxN type. Maybe push back on it and see if Vyper can just adopt the fixedMxN type? (or, even just pick one, like `fixed256x10`)

> We were talking about this, but it is hard to verify formally. So we were reluctant.

This doesn't make sense to me. Verify what?

davesque:
@carver Yeah it would be nice to get some clarification on the difficulty with formal verification.

For now, one of the devs involved in that ticket has told me that I can get away with doing some bounds checking on values as they're being decoded or encoded which seems correct to me.  I think it should be pretty easy for us to have support for both `fixed` and `decimal` types so maybe there's no need for any pushback?

carver:
> so maybe there's no need for any pushback?

I think of the proliferation of similar standards as a kind of [Public Bad](https://en.wikipedia.org/wiki/Public_bad), which forces work on all the software that supports it. The cost isn't that high for us, but there are global costs to having yet another type. The global benefit/cost seem poor in this case. But fighting the proliferation is personally high-cost and very-low-benefit, so whether it's worth it is a personal decision of how charitable I'm feeling. :) One potential benefit of pushing back is that it helps me learn the subtleties of undocumented differences.

> For now, one of the devs involved in that ticket has told me that I can get away with doing some bounds checking on values as they're being decoded or encoded which seems correct to me.

Like this -- deserves a lot more documentation if it's to be a proper standard.

davesque:
> deserves a lot more documentation if it's to be a proper standard.

Yep, I agree.  There's now an open issue for this: https://github.com/ethereum/solidity/issues/3869

davesque:
@carver @pipermerriam Mkay.  The fixed types should be ready to merge.  I guess we just need to figure out what we're going to do about decimal types.

eth_abi/decoding.py
```diff
@@ -340,6 +341,74 @@ def from_type_str(cls, abi_type, registry):
...
+    is_big_endian = True
+
+    @classmethod
+    def validate(cls):
+        super().validate()
+
+        if cls.frac_places is None:
+            raise ValueError("must specify `frac_places`")
+
+        if not (cls.frac_places > 0 or cls.frac_places <= 80):
```
pipermerriam: do you find the following easier to parse?

```python
if frac_places <= 0 or frac_places > 80:  # I like this one more because it doesn't have a `not` in it.
    pass
if not 0 < frac_places <= 80:
    pass

File: eth_abi/decoding.py [hunk 2]
```diff
@@ -340,6 +341,74 @@ def from_type_str(cls, abi_type, registry):
...
+            raise ValueError("`frac_places` must be in range (0, 80]")
+
+
+class UnsignedFixedDecoder(BaseFixedDecoder):
+    @classmethod
+    def decoder_fn(cls, data):
+        value = big_endian_to_int(data)
+
+        with decimal.localcontext(abi_decimal_context):
+            decimal_value = decimal.Decimal(value) * TEN ** -cls.frac_places
```
pipermerriam: I know they are equal, but I expected this to be `decimal.Decimal(value) / TEN ** cls.frac_places`.  Anyone else find that more intuitive or semantically correct?

pipermerriam (approved):

eth_abi/decoding.py
carver: Yes, that's also what I would have expected. It's a bit more visually apparent what's happening. Although some of it is that I don't read `-variable` as smoothly as I read `-1 * variable`.

eth_abi/decoding.py
carver: :+1: to:

```py
if frac_places <= 0 or frac_places > 80:
    raise
```

---

Also, wasn't there some requirement about `bit-size % 8 == 0`?

eth_abi/decoding.py
```diff
@@ -340,6 +341,74 @@ def from_type_str(cls, abi_type, registry):
...
+
+        return decimal_value
+
+    @parse_type_str('ufixed')
+    def from_type_str(cls, abi_type, registry):
+        value_bit_size, frac_places = abi_type.sub
```
carver, line 371: What does the error look like if someone does something weird like `ufixed8`?

eth_abi/encoding.py
```diff
@@ -269,6 +284,88 @@ def from_type_str(cls, abi_type, registry):
...
+    def bounds_fn(cls, value_bit_size):
+        return compute_signed_fixed_bounds(cls.value_bit_size, cls.frac_places)
+
+    @classmethod
+    def encode_fn(cls, value):
+        with decimal.localcontext(abi_decimal_context):
+            scaled_value = value * TEN ** cls.frac_places
+            integer_value = int(scaled_value)
+
+        unsigned_integer_value = integer_value % 2 ** cls.value_bit_size
```
carver: My preference is explicit parentheses here, but just a suggestion:

```
unsigned_integer_value = integer_value % (2 ** cls.value_bit_size)
```

(Outside of `+-*/=**`, I don't assume readers know order of operations)

carver (approved):

eth_abi/decoding.py
davesque: It seems a bit easier to go straight from the ABI spec to the if statement with not. I didn't even realize the second option you gave was valid syntax. I'm sure Guido hates it but I kinda like it in this case. 😄

I'll switch it to that.

eth_abi/encoding.py
davesque: I agree with that. That code was just lifted from one of the other coder classes so maybe I should change it there too.

eth_abi/decoding.py
davesque: @carver sorry I didn't see your comment. So you don't like the `not 0 < frac <= 80` option? The problem I have with the other option is, if you've just come from the ABI spec, it's hard to flip the greater than and less than equal in your mind. Seems like an easy way to get confused. I saw that if clause as checking whether or not an assumption laid out in the spec did not hold.

As for the modular congruence, I think that happens in the super call but I'll double check when I get back home.

davesque: Actually I just realized that code in the parens is totally wrong. It should be an `and` not an `or`.

eth_abi/decoding.py
davesque: My thinking there was that `frac_places` really just corresponds to scientific notation (e.g. `255e-2 == 2.55`).  But the ABI spec does use the notation that you prefer and that's more important.  I'll change it.

eth_abi/decoding.py
davesque: This code would throw an error: https://github.com/ethereum/eth-abi/blob/master/eth_abi/utils/grammar.py#L286-L289

The `parse_type_str` decorator parses the type and calls `ABIType.validate()`.

davesque: I don't love this approach by the way.  It would be difficult to realize that and that's one reason why I think validation should be centralized in one place (in the coder classes).

eth_abi/decoding.py
carver: I don't like `not 0 < frac <= 80` mostly because it's too many operations on one line, plus order of operations is not visibly apparent. If you really wanted to go the route of defining valid, and taking the inverse, I'd prefer:

```py
is_valid_decimal_places = (0 < frac <= 80)
if not is_valid_decimal_places:
    raise
```

These are all things that are easy to keep tweaking over time, though, so I wouldn't get too hung up on it. Except for asking the question of what new test to write to catch:

> Actually I just realized that code in the parens is totally wrong. It should be an and not an or.

PR #52, opened 2018-04-17
Move `eth_abi.utils.grammar` to `eth_abi.grammar`

davesque:
### What was wrong?

Since @carver had mentioned he might change some code to use the new `grammar.parse` function instead of `process_type`, I thought it would be a good idea to move the `grammar` module so that it lives at `eth_abi.grammar` instead of `eth_abi.utils.grammar`.

### How was it fixed?

Moved the module.  Also removed some import lint that I noticed.

#### Cute Animal Picture

Cute animal picture


eth_abi/utils/numeric.py
```diff
@@ -12,31 +10,13 @@
 )
 
 
-if sys.version_info.major == 2:
-    import struct
+def int_to_big_endian(value):
```
pipermerriam: I think we can just use the version of this from `eth-utils` and not have it in this library at all

pipermerriam (approved):

carver (approved):

PR #56, opened 2018-04-18
Make Circle CI test after merge with PR base

davesque:
### What was wrong?

With our current `config.yaml`, Circle CI does not test PRs after merging them with their respective base branches.

### How was it fixed?

Updated `config.yaml` to correct this problem.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,6 +6,22 @@ common: &common
   working_directory: ~/repo
   steps:
     - checkout
+    - run:
+        name: merge pull request base
+        command: |
+          if [[ -n "${CIRCLE_PR_NUMBER}" ]]; then
+            PR_INFO_URL=https://api.github.com/repos/$CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME/pulls/$CIRCLE_PR_NUMBER
+
+            PR_BASE_BRANCH=$(curl -L "$PR_INFO_URL" | python -c 'import json, sys; obj = json.load(sys.stdin); sys.stdout.write(obj["base"]["ref"])')
+            git fetch origin +"$PR_BASE_BRANCH":circleci/pr-base
+
+            # We need these config values or git complains when creating the
+            # merge commit
+            git config --global user.name "Circle CI"
+            git config --global user.email "circleci@ethereum.org"
+
+            git merge --no-edit circleci/pr-base
+          fi
     - restore_cache:
         keys:
           - cache-{{ .Environment.CIRCLE_JOB }}-{{ checksum "setup.py" }}-{{ checksum "tox.ini" }}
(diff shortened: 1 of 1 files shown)


davesque:
Here's a working version.  However, it requires calling out to the Github API to determine a PRs base branch.  If there's an easier way to determine the base branch, we could skip all of that.

There's also the `refs/pulls/<id>/merge` branch which Github makes available.  However, I've heard that this can be out of date if the merge fails on Github.  And I'm not aware of any way to determine if the merge failed other than to make the same request to the Github API which, worse yet, might return inaccurate information itself if the merge job on Github hasn't yet finished.

davesque:
Or, we could just always assume the base branch is `master`.

davesque:
Oh, and I should probably be using the base branch's ref and not its sha.  The sha probably just points to an ancestor commit for the PR which would never require merging.

.circleci/config.yml
```diff
@@ -6,6 +6,18 @@ common: &common
...
   steps:
     - checkout
+    - run:
+        name: merge pull request base
+        command: |
+          if [[ -n "${CIRCLE_PR_NUMBER}" ]]; then
+            PR_INFO_URL=https://api.github.com/repos/$CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME/pulls/$CIRCLE_PR_NUMBER
+            curl -L "$PR_INFO_URL" > /tmp/pr-info.json
+
+            PR_BASE_BRANCH=$(python -c 'import json, sys; obj = json.loads(open("/tmp/pr-info.json").read()); sys.stdout.write(obj["base"]["ref"])')
```
carver: Nice use of `ba-py-json`, the Tur-duck-en of scripting. :)

So `obj["base"]["ref"]` is almost always going to be master, right? I'm just trying to make sure I understand what's happening here.

.circleci/config.yml
```diff
@@ -6,6 +6,18 @@ common: &common
...
+            git fetch origin "+${PR_BASE_BRANCH}:circleci/pr-base"
+
+            git merge --no-edit circleci/pr-base
```
carver: What happens on a failure to merge here? I guess `git merge` would return an error code. Would the error code from the command cause the test to fail?

carver (approved):
Cool, looks good. After we kick the tires on this one, we can also add it to to https://github.com/carver/ethereum-python-project-template

.circleci/config.yml
davesque: As you say, the `git merge` command returns a non-zero exit status and causes the build job to fail immediately with an error message printed out by git in that build stage.  I think builds fail in general when any command returns non-zero exit status.

.circleci/config.yml
davesque: That `obj["base"]["ref"]` value is the branch name (usually "master" as you point out) for the base branch which is chosen by the author of the pull request when they create it.  As far as I know, it can be any branch in the upstream repo.

pipermerriam (approved):
Fancy! :shipit:

Issue #53, opened 2018-04-17
Use simple coder class instantiation instead of on-the-fly subclassing for particular types?


davesque:
### What was wrong?

Coder classes currently use programmatic subclassing via the built-in `type` function as an intermediate step toward creating a callable which performs the actual encoding/decoding of values:

https://github.com/ethereum/eth-abi/blob/master/eth_abi/encoding.py#L62
https://github.com/ethereum/eth-abi/blob/master/eth_abi/decoding.py#L41

### How can it be fixed?

Would it be simpler to just pass in coder settings as keyword arguments to an `__init__` method on coder classes?  So something like this:

```python
class BaseEncoder(BaseCoder):
    def __init__(self, **kwargs):
        cls = type(self)
        
        for key, value in kwargs.items():
            if not hasattr(cls, key):
                raise AttributeError(
                    'Property {key} not found on {cls_name} class. '
                    '`{cls_name}.__init__` only accepts keyword arguments which are '
                    'present on the {cls_name} class.'.format(
                        key=repr(key),
                        cls_name=cls.__name__,
                    )
                )
            
            setattr(self, key, value)

        self.validate()

    ...
```

### Disclaimer

If there's a particular reason that this approach was chosen, then just ignore me 😄 .

davesque:
@carver @pipermerriam Any thoughts on this?

pipermerriam:
@davesque no reason I see to not pursue this approach.  Previous pattern was probably borrowed from elsewhere (maybe django class-based-views?) and I'm sure there are good other ways to to do this.

carver:
No objections @davesque

davesque:
Nice.  I think then I'll probably do this as part of the work with moving validation into the coder classes.

davesque:
Closed on accident.

PR #60, opened 2018-04-20
Simplify coder creation

davesque:
### What was wrong?

The issue was discussed in ticket #53 .  The coder creation process seems a bit complicated.

### How was it fixed?

Change coder classes to use simple instantiation instead of on-the-fly subclassing to create coder callables.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/eth_abi/decoding.py
+++ b/eth_abi/decoding.py
@@ -24,7 +24,10 @@ from eth_abi.utils.numeric import (
 
 
 class BaseDecoder(BaseCoder):
-    def decode(self, stream):
+    def decode(self, stream):  # pragma: no cover
+        """
+        Decodes the given stream of bytes into a Python value.
+        """
         raise NotImplementedError('Must implement `decode`')
 
     def __call__(self, stream):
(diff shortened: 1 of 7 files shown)


davesque:
Here's a first few steps in the direction that was discussed in the issue ticket.  Let me know if it looks decent.

eth_abi/decoding.py
```diff
@@ -24,28 +24,11 @@
...
-            name = cls.__name__
-        sub_cls = type(name, (cls,), kwargs)
-        sub_cls.validate()
-        instance = sub_cls()
-        return instance
-
-    @classmethod
-    def validate(cls):
-        pass
+    def decode(self, stream):  # pragma: no cover
```
pipermerriam: Now that we're instantiating these, maybe we should make use of `abc.ABCMeta` and `abc.abstractmethod`?

pipermerriam (approved):

carver (commented):
Looks good :+1:

eth_abi/decoding.py
davesque: Sure.  Made that change.

PR #64, opened 2018-04-24
String type can't have length

davesque:
### What was wrong?

String types were incorrectly validated for numerical sub types.

### How was it fixed?

String types now may not have a sub type.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/eth_abi/grammar.py
+++ b/eth_abi/grammar.py
@@ -263,12 +263,17 @@ class BasicType(ABIType):
         base, sub = self.base, self.sub
 
         # Check validity of string type
-        if base in ('string', 'bytes'):
+        if base == 'string':
+            if sub is not None:
+                self.invalidate('string type cannot have suffix')
+
+        # Check validity of bytes type
+        elif base == 'bytes':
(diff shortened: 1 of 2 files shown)


pipermerriam (approved):

eth_abi/grammar.py
```diff
@@ -263,12 +263,17 @@ def validate(self):
         base, sub = self.base, self.sub
 
         # Check validity of string type
-        if base in ('string', 'bytes'):
+        if base == 'string':
+            if sub is not None:
+                self.invalidate('string type cannot have suffix')
+
+        # Check validity of bytes type
+        if base == 'bytes':
```
carver: Probably `elif` here, for consistent style.

carver (approved):
Oops. The PR to change it looks super clean though, which is a nice endorsement of the structure of the code. :+1:

eth_abi/grammar.py
davesque: Whoops!  Thanks for catching that.

PR #66, opened 2018-04-26
Add/enable tuple types

davesque:
### What was wrong?

Coder classes for tuple types (`MultiEncoder` and `MultiDecoder`) mostly already exists in `eth-abi` but the registry does not handle them properly.  The `HeadTrailDecoder` also lacks support for recursive dynamic types encoded with the head/tail method as detailed in the Solidity ABI spec.

### How was it fixed?

Made some efficiency improvements to the `MultiEncoder` and `MultiDecoder` classes, renamed them to `TupleEncoder` and `TupleDecoder` respectively, and created default registrations for them.  Decoding now depends on the `ContextFramesBytesIO` class which inherits from `BytesIO` and tracks extra information during decoding to help with positioning of tail offsets in nested head segments.

#### Cute Animal Picture

Cute animal picture


eth_abi/decoding.py
```diff
@@ -41,40 +41,53 @@ class HeadTailDecoder(BaseDecoder):
...
 class MultiDecoder(BaseDecoder):
     decoders = None
 
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
+        self.decoders = tuple(
+            (
+                HeadTailDecoder(tail_decoder=decoder)
+                if isinstance(decoder, (DynamicArrayDecoder, StringDecoder))
```
pipermerriam: Wondering if this should instead be a property check like `if decoder.is_dynamic`.  Otherwise 3rd party decoders won't be able to properly integrate right?

pipermerriam (approved):

eth_abi/decoding.py
davesque: Yeah I just noticed that actually.  Furthermore, any tuple type which contains at least one dynamic type becomes, itself, a dynamic type.  Since tuples can be nested, I need to check whether or not the coders in a multi coder are dynamic tuple coders.

davesque:
Left to do:
* Add support for detecting dynamic tuple coders in multi coder classes.  Tuples can be dynamic types when they contain at least one dynamic type.  @pipermerriam suggested adding some kind of `is_dynamic` API which seems like the way to go since that might be required to allow support for user-created dynamic types.
* Possibly rename `MultiEncoder` and `MultiDecoder` to `TupleEncoder` and `TupleDecoder` respectively to clarify their intended use
* Add tuple registrations to registry
* Extend/improve tests as needed

davesque:
Left to do:

* Improve existing tests if needed
* Add tests for deeply nested types, particularly nested dynamic types
* Add tests for encoding/decoding unusual edge case types such as ones like `((),(((),(),()),()),())` which should encode to and from an empty byte string

davesque:
Had to do a little bit of a weird thing.  I made a wrapper class for `BytesIO` which tracks some contextual information during decoding of byte streams.  This, or something like it, is necessary since tuples only calculate tail offsets relative to their own encoding, regardless of whether or not they appear in some other nested type structure.  So you need to manage some kind of state during decoding to keep track of the relative meaning of the offsets.

davesque:
Created issue for lack of support for non-standard packed mode: #67

davesque:
@pipermerriam @carver Now that we're getting proper tuple support via `encode_single('(int,int,...)', val)` and similarly with `decode_single`, would it make sense to remove the `encode_abi` and `decode_abi` functions altogether and rename `encode_single` and `decode_single` to `encode` and `decode`?

pipermerriam:
Yes, but only through a proper deprecation cycle.

- introduce new API
- add deprecation warnings to old API
- drop old API in major version bump

Given that web3 uses `eth-abi` we should probably plan to backport any bugfixes to the `1.x` line for the next 6-12 months.

eth_abi/decoding.py
```diff
@@ -24,6 +27,47 @@
 )
 
 
+class ContextFramesBytesIO(io.BytesIO):
+    """
+    A byte stream which can track a series of contextual frames in a stack.
+    """
```
pipermerriam: This *might* be easier to understand with a diagram?

```
index: 00 01 02 03 04 05 06 07
bytes: de ad be ef 31 41 59 26

<something here to diagram a little what a frame is...>
```

eth_abi/decoding.py
```diff
@@ -24,6 +27,47 @@
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self._frames = [(0, 0)]
```
pipermerriam: I don't understand this initial value.  Looking at the code there doesn't appear to be a code path that can even touch `self._frames[0]`?

pipermerriam (approved):

eth_abi/decoding.py
davesque: Yep, you're right.  I need to take that out.  I had it that way because, previously, I was referencing the current context frame by doing `self._frames[-1]`.  But then I began using `self._total_offset` instead and forgot to reconsider the use of a default first frame.

.circleci/config.yml
```diff
@@ -24,20 +24,20 @@ common: &common
...
         keys:
-          - v1-cache-{{ .Environment.CIRCLE_JOB }}-{{ checksum "setup.py" }}-{{ checksum "tox.ini" }}
+          - v2-cache-{{ .Environment.CIRCLE_JOB }}-{{ checksum "setup.py" }}-{{ checksum "tox.ini" }}
     - run:
         name: install dependencies
         command: pip install --user tox
     - run:
         name: run tox
-        command: ~/.local/bin/tox
+        command: ~/.local/bin/tox -- -n 8
```
pipermerriam: FYI, environment variables work well for this too.  `PYTEST_ADDOPTS="-n 2"` (though you have to explicitely whitelist the environment variable in tox)

Also, the ci-environment only has 2 cores so you should probably keep the concurrency down to around 2-4 max unless you know that test running is not CPU bound.

davesque: Oh, where did you see that?  When I looked at `/proc/cpuinfo` it was claiming there were 8 cores.

davesque: I see now that the "Resources" printout in the Circle CI interface is saying that the jobs are given 2 CPUs on which to run like you said.  I'm seeing speed gains until I go past 4 workers.  I'll just keep it at `-n 4` for now.

pipermerriam: 👍 somewhere in the docs. I can't remember exactly where.  Probably somewhere in the container or parallelization stuff.

davesque:
@pipermerriam Alright, this is as done as done can be.  Ready to merge this.

To keep you in the loop, most of the commits since your last comment re-organized the test modules to more closely mirror the structure of the package modules.  My hope is that this will make it easier to understand at a glance what parts of the package are tested by different parts of the test suite.

Also, I added explicit tests for a decent range of tuple encodings/decodings and also reversibility tests for tuple values.

davesque:
@pipermerriam Alright, this is as done as done can be. Ready to merge this.

To keep you in the loop, most of the commits since your last comment re-organized the test modules to more closely mirror the structure of the package modules. My hope is that this will make it easier to understand at a glance what parts of the package are tested by different parts of the test suite.

Also, I added explicit tests for a decent range of tuple encodings/decodings and also reversibility tests for tuple values.

PR #68, opened 2018-05-08
Add abi.is_encodable

davesque:
### What was wrong?

Currently, there is a poorly located and maintained version of `is_encodable` here: https://github.com/ethereum/web3.py/blob/master/web3/utils/abi.py#L148

This should be a part of the `eth_abi` library and should work in conjunction with new features like tuples and the registry.

### How was it fixed?

Made `validate_value` part of public API for encoder classes and added implementations of it where necessary.  Also added `abi.is_encodable` which is implemented via that new functionality.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/eth_abi/encoding.py
+++ b/eth_abi/encoding.py
@@ -53,12 +53,11 @@ class BaseEncoder(BaseCoder, metaclass=abc.ABCMeta):
         """
         pass
 
+    @abc.abstractmethod
     def validate_value(self, value):
         """
         Determines whether or not the given value can be encoded by this
-        encoder.  Encoder classes can implement validation of values if desired.
-        By default, this method does nothing and an encoder will attempt to
-        encode any value it is given.
+        encoder.
         """
         pass
 
--- a/eth_abi/encoding.py
+++ b/eth_abi/encoding.py
@@ -522,11 +522,12 @@ class ByteStringEncoder(BaseEncoder):
     def encode(cls, value):
         cls.validate_value(value)
 
-        encoded_size = encode_uint_256(len(value))
         if not value:
             padded_value = b'\x00' * 32
         else:
             padded_value = zpad_right(value, ceil32(len(value)))
+
+        encoded_size = encode_uint_256(len(value))
         encoded_value = encoded_size + padded_value
 
         return encoded_value
@@ -555,11 +556,12 @@ class TextStringEncoder(BaseEncoder):
 
         value_as_bytes = codecs.encode(value, 'utf8')
 
-        encoded_size = encode_uint_256(len(value_as_bytes))
         if not value_as_bytes:
             padded_value = b'\x00' * 32
         else:
             padded_value = zpad_right(value_as_bytes, ceil32(len(value_as_bytes)))
+
+        encoded_size = encode_uint_256(len(value_as_bytes))
         encoded_value = encoded_size + padded_value
 
         return encoded_value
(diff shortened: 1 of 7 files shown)


davesque:
Still need to add some basic tests for this.

davesque:
Differences from `web3.utils.abi.is_encodable`:

* Does not check whether or not an address value is an ENS name
* Does not allow `str` values (hex strings) for `bytes` types

pipermerriam:
Both of the `web3` differences are things we can handle on the web3 side as a normalization step.

eth_abi/encoding.py
```diff
@@ -53,6 +53,15 @@ def encode(self, value):  # pragma: no cover
...
         pass
 
+    def validate_value(self, value):
+        """
+        Determines whether or not the given value can be encoded by this
+        encoder.  Encoder classes can implement validation of values if desired.
+        By default, this method does nothing and an encoder will attempt to
+        encode any value it is given.
+        """
+        pass
```
pipermerriam: Hrm, having values pass validation by default seems like a foot gun.  Makes it really easy to accidentally write an encoder which will always pass a check for `is_encodable` but otherwise won't work with the given value.

Maybe I'm missing something, but I'd suggest going with either default implementation raising an error, or making this an `abstractmethod` to require subclasses to implement it. (preference is towards the later `abstractmethod` approach).

eth_abi/encoding.py
```diff
@@ -512,18 +537,33 @@ def from_type_str(cls, abi_type, registry):
...
             )
+
+    @classmethod
+    def encode(cls, value):
+        cls.validate_value(value)
+
         value_as_bytes = codecs.encode(value, 'utf8')
-        return super().encode(value_as_bytes)
+
+        encoded_size = encode_uint_256(len(value_as_bytes))
```
pipermerriam: *nitpick*

Can we move this line down below the `if/else` statement so that it's directly adjacent to line 564 where it's used.

tests/test_abi/test_is_encodable.py
```diff
@@ -0,0 +1,64 @@
...
+        ('int', True),
+        ('uint8', 256),
+        ('uint8', -1),
+        ('int8', 128),
+        ('int8', -129),
+        ('string', -129),
+        ('bytes', 129),
+        ('bytes5', b'asdfgt'),
+        ('address', '0x8deebe59332ebc3f81935e311f0fb8556cf573'),
+        ('address', decode_hex('0x8deebe59332ebc3f81935e311f0fb8556cf573')),
```
pipermerriam: Why does this one fail?

pipermerriam (approved):

eth_abi/encoding.py
davesque: I mostly did it this way for expedience since the casper people need something quickly and there were some encoders that didn't validate their input values.  So it didn't seem like we'd break the behavior of the library significantly by allowing a default no-op.  But actually I think there are no encoders at this point which lack a `validate_value` method so perhaps we can just make it an abstract method like you say.

tests/test_abi/test_is_encodable.py
davesque: If you mean the `('bytes', 129)` line, it's because it's not a `bytes` or `bytearray` value.

davesque: If you mean the `('address', ...)` line, it's because the address isn't 20 bytes long.  I'll add comments for both cases.

carver (approved):
:+1: good to go

At some point, I want to come back to this, to make sure that encoding performance didn't drop significantly in pypy. (that the slowdown was only in test example generation)

PR #70, opened 2018-05-08
Readme updates

davesque:
### What was wrong?

The readme was a bit out of date and contained incorrect formatting directives.

### How was it fixed?

Cleaned up the formatting and resolved any inconsistencies with current functionality.

#### Cute Animal Picture

Cute animal picture


README.md
```diff
@@ -1,104 +1,119 @@
...
 pip install eth-abi
 ```
 
+## Decoding
 
+Decoding functions convert ABI-encoded binary data generated by the EVM into
+python values.
 
-## Documentation
+### eth\_abi.decode\_single(typ, data)
```
pipermerriam, line 27: I would wrap these titles in backticks to make the format as code blocks.  I think they are easier to read that way.

pipermerriam (approved):

pipermerriam:
Oh, it might be worth listing the supported ABI types?

README.md
davesque: The lack of backticks was on purpose actually.  The backticks didn't look great IMHO :(.  Also, the formatting was a bit nicer too when converted to rst.

davesque:
I do plan to go through and update the sphinx documentation to include more detailed information like that and also details about new features like the registry.

carver:
IMO, we shouldn't have any user-facing docs in the readme at all, they should be directed to RTD, instead. The readme should only be for people who want to contribute to the project. So a number of the things in the readme now should be moved into RTD.

davesque:
Actually, I began going through the sphinx docs last night and started feeling the same way as you, @carver .  I think I might just rebuild this pull request to that effect.

PR #71, opened 2018-05-09
IllegalValue if fixed value has too many decimals

carver:
### What was wrong?

`is_encodable` was crashing when a decimal has too many digits to encode with the given fixed ABI type, because it was raising `ValueError` which isn't caught.

### How was it fixed?

Raise `IllegalValue` instead

#### Cute Animal Picture

Cute animal picture


pipermerriam (approved):

carver:
Thanks @pipermerriam !

davesque (approved):
Looks good to me.

PR #72, opened 2018-05-09
disallow encoding floats to fixed type

carver:
### What was wrong?

`float`s crashed when trying to encode them, because of the check for decimal "residue". They are imprecise anyway, by nature, and the eth-abi must be precise.

### How was it fixed?

Disallow `float` as a valid value to encode to a `fixedMxN` type.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/eth_abi/encoding.py
+++ b/eth_abi/encoding.py
@@ -291,7 +291,10 @@ class SignedIntegerEncoder(NumberEncoder):
 
 class BaseFixedEncoder(NumberEncoder):
     frac_places = None
-    type_check_fn = staticmethod(is_number)
+
+    @staticmethod
+    def type_check_fn(value):
+        return is_number(value) and not isinstance(value, float)
 
     @staticmethod
     def illegal_value_fn(value):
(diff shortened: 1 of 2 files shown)


pipermerriam:
I don't think we have to disallow them (thought I'm not opposed to it).  The `eth-utils.to_wei` API accepts floats by using their string representations.

https://github.com/ethereum/eth-utils/blob/master/eth_utils/currency.py#L84-L85

I *think* this is a viable option, but I am not sure there aren't edge cases where *special* floats will evaluate to unexpected values.

pipermerriam:
link: https://github.com/ethereum/eth-utils/issues/102

davesque:
Personally, I like the idea of disallowing them.  That way, we have a better idea of the expected behavior on our end.  Also, people who want to use them can figure out a reliable way to convert to decimal for their purposes (whether it be via the string representation or something else).  I don't like the idea of encouraging people to use floats in the context of the Ethereum blockchain.

davesque (approved):

carver:
Right, the string representation works nicely for float literals, which I think is a common use case for `to_wei`. I'd be worried that a common case here is having the value be a result of a computation, when the string representation won't save you from rounding errors accumulated during computation.

Also, it's much easier to change our minds to add it than it would be to take it away, so deciding to merge this is half-opinion, half-punt.

PR #73, opened 2018-05-25
Add documentation and some type hints

davesque:
### What was wrong?

Documentation was a bit out of date in various ways.  Also needed documentation for registry feature.

### How was it fixed?

Added API documentation via docstrings.  Updated existing documentation.  Added registry section.  Added informative type hints.

#### Cute Animal Picture

Cute animal picture


davesque:
Actually, sorry.  This is not yet ready for review.  I requested that prematurely.

eth_abi/abi.py
```diff
@@ -1,28 +1,44 @@
...
 
-from eth_abi.exceptions import EncodingError
-
 from eth_abi.encoding import TupleEncoder
 
+from eth_abi.exceptions import EncodingError
+
 from eth_abi.registry import registry
 
+from eth_abi.typing import (
```
pipermerriam: Can you move these to a separate python module called `eth_typing`.  Taking a look in the py-evm repo for how it's done there.  We're planning on extracting these to a common repo that can be shared across our codebases.

I think this should make migrating to this new repo once it's been created easier.

davesque: Sure, no problem.

davesque: So that would be a separate module which is also installed by setup.py when installing eth-abi?  It doesn't seem like the installer would gracefully manage merging different versions of that module into the import hierarchy.  For example, wouldn't my version of `eth_typing/__init__.py` differ from py-evm's version and get overwritten depending on the order in which things are installed?

davesque:
@pipermerriam Okay, this is ready for review now.

pipermerriam (approved):

Makefile
```diff
@@ -31,7 +31,7 @@ test-all:
 	tox
 
 build-docs:
-	sphinx-apidoc -o docs/ . setup.py "eth_abi/utils/*" "*conftest*" "tests/*"
+	sphinx-apidoc -o docs/ . setup.py 'eth_typing/*' 'eth_abi/encoding.py' 'eth_abi/decoding.py' 'eth_abi/utils/*' 'tests/*'
```
carver: Why not generate API documentation for encoding & decoding?

eth_typing/abi.py
```diff
@@ -0,0 +1,4 @@
+from typing import Union
+
+TypeStr = str
+Decodable = Union[bytes, bytearray]
```
carver: I agree with the direction you went in chat: moving this module to an external library and adding the dependency.

Makefile
davesque: I guess we don't need to exclude those modules.  I think I did that because I was trying to find a way to exclude certain things from being included in the API documentation.  But then I realized I could restrict things from showing up with the `:members:` option on the `automodule` directive.  I think I can revert this change.

carver (approved):

PR #74, opened 2018-05-31
Add type hints

### LINKED ISSUE #33: Add type hints and mypy to CI
### What was wrong?

We should add type hints to the code as well as a `mypy` run to the CI environment.

### How can it be fixed?

Copy implementation from `eth-keys` or `py-evm` repo.

palash25:
Also tox.ini: Add mypy to the CI

Closes: https://github.com/ethereum/eth-abi/issues/33

Sorry for taking so long to open this PR. I was busy with exams and GSoC preparations.
I will keep updating this.

### What was wrong?
Type hints were not present

### How was it fixed?
Mypy was added to the CI and type hints added to the packages

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/eth_abi/utils/numeric.py
+++ b/eth_abi/utils/numeric.py
@@ -1,4 +1,7 @@
 import decimal
+from typing import (
+    Tuple,
+)
 
 abi_decimal_context = decimal.Context(prec=999)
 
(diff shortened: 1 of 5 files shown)


eth_abi/utils/padding.py
```diff
@@ -4,23 +4,23 @@
 
 
 @curry
-def zpad(value, length):
+def zpad(value:str, length:int) -> str:
```
pipermerriam: Can you add spaces after the `:` for style consistency.

palash25: Oh sure. I was going to lint it after I was done with it. I will do that before asking for a review its still WIP. :smile:

eth_abi/utils/numeric.py
```diff
@@ -46,7 +47,7 @@ def compute_signed_fixed_bounds(num_bits, frac_places):
     return lower, upper
 
 
-def compute_unsigned_real_bounds(num_high_bits, num_low_bits):
+def compute_unsigned_real_bounds(num_high_bits: int, num_low_bits: int) -> Tuple[int, int]:
```
davesque: Both `compute_unsigned_real_bounds` and `compute_signed_real_bounds` should return `Tuple[float, float]` if I'm not mistaken.

eth_abi/utils/numeric.py
```diff
@@ -66,7 +67,7 @@ def compute_signed_real_bounds(num_high_bits, num_low_bits):
     )
 
 
-def quantize_value(value, decimal_bit_size):
+def quantize_value(value: int, decimal_bit_size: int) -> decimal.Decimal:
```
davesque: This function is doing rounding on a value which is likely to have a fractional component.  So it probably won't ever get an `int` as an input for `value`.  I'm pretty sure value's type should be `Union[float, decimal.Decimal]` or something like that.

tox.ini
```diff
@@ -9,15 +9,20 @@ exclude= tests/*
...
 [testenv]
 commands=py.test --tb native {posargs:tests}
-deps=
+deps =
     -r{toxinidir}/requirements-dev.txt
 basepython=
   py35: python3.5
   py36: python3.6
   pypy3: pypy3
-passenv=PYTEST_ADDOPTS
```
davesque: This line shouldn't have been removed.

eth_abi/utils/padding.py
davesque: Also, I'm pretty sure that this function, and the others for which type annotations were added, accept `bytes` values and not `str`.

davesque (approved):

PR #76, opened 2018-06-05
Fix support for callable coders

davesque:
### What was wrong?

When adding support for tuple types and the `is_encodable` function, support for simple callable coders was broken.

### How was it fixed?

Modified code in `is_encodable` as well as `encoding` and `decoding` modules to handle simple callables.  Added tests for functionality that wasn't working as expected.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/eth_abi/base.py
+++ b/eth_abi/base.py
@@ -144,4 +144,4 @@ class BaseCoder:
         Used by ``ABIRegistry`` to get an appropriate encoder or decoder
         instance for the given type string and type registry.
         """
-        raise NotImplementedError('Must implement `from_type_str`')
+        raise cls()
--- a/tests/test_integration/test_custom_registrations.py
+++ b/tests/test_integration/test_custom_registrations.py
@@ -1,0 +1,99 @@
+from eth_abi.abi import (
+    decode_single,
+    encode_single,
+)
+from eth_abi.decoding import (
+    BaseDecoder,
+)
+from eth_abi.encoding import (
+    BaseEncoder,
+)
+from eth_abi.exceptions import (
+    DecodingError,
+    EncodingError,
+)
+from eth_abi.registry import (
+    registry,
+)
+
+NULL_ENCODING = b'\x00' * 32
+
+
+def encode_null(x):
+    if x is not None:
+        raise EncodingError('Unsupported value')
+
+    return NULL_ENCODING
+
+
+def decode_null(stream):
+    if stream.read(32) != NULL_ENCODING:
+        raise DecodingError('Not enough data or wrong data')
+
+    return None
+
+
+class EncodeNull(BaseEncoder):
+    word_width = None
+
+    @classmethod
+    def from_type_str(cls, type_str, registry):
+        word_width = int(type_str[4:])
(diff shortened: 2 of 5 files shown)


tests/test_integration/test_custom_registrations.py
```diff
@@ -0,0 +1,99 @@
...
+
+    assert encode_single('null', None) == NULL_ENCODING
+    assert decode_single('null', NULL_ENCODING) is None
+
+    encoded_tuple = encode_single('(int,null)', (1, None))
+
+    assert encoded_tuple == b'\x00' * 31 + b'\x01' + NULL_ENCODING
+    assert decode_single('(int,null)', encoded_tuple) == (1, None)
+
+    registry.unregister('null')
```
carver, line 80: This `unregister` being necessary in the test is a hint to me that the API could be improved. The ideal I would look for is the pattern of initializing an object in a fixture, and then not worrying about state from one test polluting another test. That's awkward with this global `registry` singleton.  Something like a `make_registry()` method that initializes a new registry with all the defaults would enables us to do:

```py
@pytest.fixture
def registry():
    return make_registry()
```

Then we don't have to worry about unregistering.

Since there are other more pressing issues, maybe just capture the idea into an issue and we can come back to it later. (if you like it)

carver (approved):

tests/test_integration/test_custom_registrations.py
davesque: I thought about making a fixture but just decided it was an okay hack for the time being.  Also, I felt like it would be useful to test the functionality by calling `encode_single` and `decode_single` to make it more like an integration test instead of querying the registry directly and using the result of that.  I'll make an issue for it.

davesque: Issue created: #77

eth_abi/base.py
```diff
@@ -144,4 +144,4 @@ def from_type_str(cls, type_str, registry):  # pragma: no cover
         Used by ``ABIRegistry`` to get an appropriate encoder or decoder
         instance for the given type string and type registry.
         """
-        raise NotImplementedError('Must implement `from_type_str`')
+        raise cls()
```
pipermerriam, line 147: This line is confusing to me.

tests/test_integration/test_custom_registrations.py
pipermerriam: This should be wrapped in a `try/finally`, otherwise it's going to leave the registry polluted when this test fails.

eth_abi/base.py
davesque: Thank you for catching that.  I intended for that to be `return cls()` and I believe my thinking was that it would make some of the examples in the registry documentation more simple but then I just went with slightly more complex examples.  I'll make a PR to revert it right away.

davesque: PR created for this: #78

tests/test_integration/test_custom_registrations.py
davesque: PR created for this: #79

PR #78, opened 2018-06-05
Fix typo introduced in PR #76

davesque:
### What was wrong?

A typo was introduced in PR #76 .

### How was it fixed?

Reverted the change.

#### Cute Animal Picture

Cute animal picture


davesque:
This is just a one-liner so I'm just gonna merge this.

Issue #81, opened 2018-07-12
decode_abi with a string returns a bytes value


carver:
* Version: df3fc70cc0 (master at creation time)
* Python: 3.6
* OS: linux

### What was wrong?

I would expect that decoding an abi value of `string` would return a `str`-typed value.

```py
In [1]: from eth_abi import decode_abi

In [2]: decode_abi(['string'], b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00 \x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x05mystr\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00')
Out[2]: (b'mystr',)
```
I expected `('mystr',)`

### How can it be fixed?

Last step of decoding should be to call `.decode('utf8')` on a bytes value if using a string decoder.

pipermerriam:
Noting that this will result in `string` values that cannot be decoded due to containing invalid UTF8 characters, but I think this is *ok* since the [docs for the `string` type](https://solidity.readthedocs.io/en/develop/abi-spec.html#types) state that it is *assumed to be UTF8 encoded*.

carver:
> assumed to be UTF8 encoded

Right, although it might be desirable for users to retrieve the `bytes` data even if it's not valid utf8. (rather than just crash)

*Maybe* would should use something like `.decode("utf-8", "replace")`. Although I kind of hate it already, the idea that broken strings won't be noticed early on. I'm just not sure how else to reasonably give people access to malformed data. Maybe we just don't. If you want malformed data, you write a curl call to the json-rpc.

I guess I landed on: crash if not valid utf8.

pipermerriam:
How about we raise a custom error message that is as informative as possible when this happens.

> The returned type for this function is `string` which is expected to be a UTF8 encoded string of text.  The returned value ... could not be decoded as valid UTF8.  This is indicative of a broken application which is using incorrect return types for binary data.  <something about how to parse it manually>

dylanjw:
[...]

PR #82, opened 2018-07-16
Return string type as utf-8 decoded string

dylanjw:
### What was wrong?
string types are return as utf-8 encoded bytes.  Makes more sense to return as utf-8 decoded string type.

See #81 

### How was it fixed?
Added a utf-8 decoding step to the StringDecoder.

#### Cute Animal Picture
image

### DIFF
--- a/eth_abi/decoding.py
+++ b/eth_abi/decoding.py
@@ -608,11 +608,15 @@ class StringDecoder(ByteStringDecoder):
     def decoder_fn(data):
         try:
             value = data.decode("utf-8")
-        except UnicodeDecodeError:
+        except UnicodeDecodeError as e:
             raise UnicodeDecodeError(
+                e.encoding,
+                e.object,
+                e.start,
+                e.end,
                 "The returned type for this function is string which is "
                 "expected to be a UTF8 encoded string of text. The returned "
-                "value {0} could not be decoded as valid UTF8. This is indicative "
+                "value could not be decoded as valid UTF8. This is indicative "
                 "of a broken application which is using incorrect return types for "
-                "binary data.".format(data))
+                "binary data.")
         return value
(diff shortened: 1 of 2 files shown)


dylanjw:
Im raising a `UnicodeDecodeError`.  Should I replace that with a custom exception?

eth_abi/decoding.py
```diff
@@ -594,12 +594,29 @@ def read_data_from_stream(stream):
...
+    @parse_type_str('string')
     def from_type_str(cls, abi_type, registry):
         return cls()
+
+    @staticmethod
+    def decoder_fn(data):
+        try:
+            value = data.decode("utf-8")
+        except UnicodeDecodeError as e:
+            raise UnicodeDecodeError(
```
pipermerriam: This should probably just be the local `DecodingError` from this library.

carver: Also, `raise ... from e` will keep all the original decoding error details in the stack.

tests/test_decoding/test_decoder_properties.py
```diff
@@ -183,24 +185,87 @@ def test_decode_signed_int(integer_bit_size, stream_bytes, data_byte_size):
...
+)
+def test_decode_strings_raises(_bytes, pad_size):
+    size_bytes = zpad32(int_to_big_endian(len(_bytes)))
+    padded_bytes = _bytes + b'\x00' * pad_size
+    stream_bytes = size_bytes + padded_bytes
+    stream = ContextFramesBytesIO(stream_bytes)
+
+    decoder = StringDecoder()
+
+    st.assume(not is_utf8_decodable(_bytes))
```
pipermerriam: Instead of this you can do a `.filter()` call on the bytes strategy.

```python
from cytoolz import compliment

@given(_bytes=st.binary(...).filter(compliment(is_utf8_decodable)), ...)
```

This should only give you byte strings which are not utf8 decodable.

pipermerriam (approved):

Issue #85, opened 2018-07-26
Remove `process_type` utility function in favor of `grammar.parse`


davesque:
### What was wrong?

The `process_type` utility function (found [here](https://github.com/ethereum/eth-abi/blob/master/eth_abi/utils/parsing.py#L11)) is left over from the previous style of parsing type strings.  The `parse` function in the `grammar` module (found [here](https://github.com/ethereum/eth-abi/blob/master/eth_abi/grammar.py#L357)) provides the same functionality and is more extensible.

### How can it be fixed?

We should remove the `process_type` function in favor of the `grammar.parse` function.

carver:
It's used a few places in web3, so that's one place to try out the new approach.

davesque:
@carver My thinking was that this could be something that gets included in v2.

pipermerriam:
If web3 is using this we need to remove that since this API is under `utils` and thus, is considered private.  I don't think this requires a deprecation, but we should be careful to not break web3 as we fix this.

davesque:
@pipermerriam @carver 

Another behavior that I'd like to change, which is related to all of this, is that the `encode_single` and `decode_single` functions accept either a type string or a type tuple (with "base", "sub", and "arrlist" components).  If a type tuple is provided, the `collapse_type` utility function is used to convert it to a type string.

I'd like to remove the `collapse_type` function and make it so that the `encode_single` and `decode_single` functions expect a type string and nothing else.  My motivation with this is to make the API simpler and less magical.

I feel like I might have had a previous discussion about this with @carver in which he may have asked to leave this behavior unchanged.  Sorry if I'm forgetting this.  If that is the case, would it be possible to simply move the `collapse_type` function into web3 or wherever it's needed and apply it manually to type string data that is headed for `encode_single` or `decode_single`?

davesque:
^^^ Actually, I'll make another issue ticket for the last comment.

davesque:
I went ahead and made PR #87 for when we decide to include this.

dylanjw:
I can open a PR in web3 to replace `process_type` with `grammar.parse`

davesque:
Merged PR #87 .

PR #87, opened 2018-07-27
Remove process_type utility function

davesque:
### What was wrong?

See issue #85 .

### How was it fixed?

Removed `process_type`.

#### Cute Animal Picture

Cute animal picture


eth_abi/utils/parsing.py
```diff
@@ -1,48 +1,6 @@
 """
 This module provided for backwards compatibility.
 """
-from eth_abi.grammar import (
-    TupleType,
```
pipermerriam: I must have missed this when it got merged, but worth pointing out now.

The `./project/utils` module in our repos has some conventions that help keep those modules well isolated.

- everything in `utils` is private API
- anything in `utils` may not import from outside of utils.

If you have to break the import rule which occasionally happens, the imports should happen inline in the function bodies to prevent circular imports from happening.

pipermerriam (approved):

eth_abi/utils/parsing.py
davesque: Nice.  I'll keep that in mind from here on out.

tests/test_utils/test_parsing.py
```diff
@@ -1,83 +1,21 @@
...
-    'typestr, expected_parse',
-    (
-        ('uint256', ('uint', '256', [])),
-        ('uint', ('uint', '256', [])),
-        ('uint256[]', ('uint', '256', [[]])),
-        ('uint[]', ('uint', '256', [[]])),
-        ('int[2]', ('int', '256', [[2]])),
-        ('function', ('bytes', '24', [])),
-        ('fixed', ('fixed', '128x18', [])),
-        ('ufixed', ('ufixed', '128x18', [])),
```
carver, line 21: Looks like these last three cases aren't tested by the grammar parser https://github.com/ethereum/eth-abi/blob/master/tests/test_grammar.py#L77

Maybe now is a good time to add them.

davesque: Added some tests in the `test_grammar` module for those cases.

carver (approved):

Issue #67, opened 2018-04-28
Add support for non-standard packed mode


davesque:
### What was wrong?

We should add support for non-standard packed mode as described in the solidity ABI spec here: https://solidity.readthedocs.io/en/develop/abi-spec.html#non-standard-packed-mode

### How can it be fixed?

I imagine support could be added in the form of another encoding function which could be imported by doing `from eth_abi import encode_packed` or something like it.  There is no need for a corresponding decoding function since the format is ambiguous and not parsable.

The `encode_packed` function would probably need to use a separate `packed_registry` as well as separate packed coder classes which would be routed by that registry.

Uxio0:
Would be very useful now that `abi.encode_packed` will be mandatory in solidity before calling keccak on data.

relyt29:
Is this pushed to Pypi? I can't seem to use it from the pip package (1.2.2)

Edit: Looks like this is released in the 2.0 version, but the pre-releases are on pypi. Guess I'll be using those then >.<

PR #89, opened 2018-07-28
Add support for non-standard packed mode

davesque:
### What was wrong?

See issue #67 .

### How was it fixed?

Added `registry_packed` registry for packed mode encoders.  Added packed mode encoder classes.  Added `encode_single_packed` and `encode_abi_packed` functions.

#### Cute Animal Picture

Cute animal picture


pipermerriam (approved):
My eyes admittedly glazed over while perusing the tests.  It looks a lot like packed encodings were easy to implement within the existing encoder/decoder framework

carver (approved):
LGTM

davesque:
@pipermerriam Yeah, it was pretty straight forward.  The tests were the hardest point as you allude to.  The thing that probably seems most confusing is that little utility function `words`.  To keep you in the loop, my goals with creating that were:

* To have a way to describe the expected results of encodings without using Python's very verbose literal byte syntax
* To avoid having to literally spell out all padding bytes (of which there are many with 32 byte wide words)
* To be able to use more or less the same encoding "description" for both standard and packed mode

davesque:
Ugh.  Accidentally clicked the close button.  I wish Github had a confirmation dialogue for that or something :/.

PR #92, opened 2018-08-14
Update arrays to use head-tail encoding

### LINKED ISSUE #93: Support for arrays of variable length types is broken
Here's the canonical way (per Solidity, at least) to encode `(bytes[]("0xf8a8fd6d", "0xf8a8fd6d"))`:

```
0000000000000000000000000000000000000000000000000000000000000020
0000000000000000000000000000000000000000000000000000000000000002
0000000000000000000000000000000000000000000000000000000000000040
0000000000000000000000000000000000000000000000000000000000000080
0000000000000000000000000000000000000000000000000000000000000004
f8a8fd6d00000000000000000000000000000000000000000000000000000000
0000000000000000000000000000000000000000000000000000000000000004
f8a8fd6d00000000000000000000000000000000000000000000000000000000
```

Broken down:

 - 0x00: A head component, consisting of an offset for the tail data (0x20)
 - 0x20: The tail of the array, specifying its length (2)
 - 0x40: The head component of the first element of the array: the offset for the tail data, relative to the start of the current tail data (0x40 + 0x40 = 0x80)
 - 0x60: The head component of the first element of the array: the offset for the tail data, relative to the start of the current tail data (0x80 + 0x40 = 0xC0)
 - 0x80: The length of the first array element (4)
 - 0xA0: The first array element (0xf8a8fd6d)
 - 0xC0: The length of the second array element (4)
 - 0xE0: The second array element (0xf8a8fd6d)

This is most easily thought of as a recursive encoding: The tail data for the first (and in this case, only) element of the ABI is the entire ABI encoding of that element.

In contrast, this library presently produces the current output:

```
0000000000000000000000000000000000000000000000000000000000000020
0000000000000000000000000000000000000000000000000000000000000002
0000000000000000000000000000000000000000000000000000000000000004
f8a8fd6d00000000000000000000000000000000000000000000000000000000
0000000000000000000000000000000000000000000000000000000000000004
f8a8fd6d00000000000000000000000000000000000000000000000000000000
```

The head element, and the length of the array are correct, but then array elements are packed tightly, without the correct head/tail recursive encoding.

davesque:
### What was wrong?

Array encoding was not using head-tail encoding when necessary (for dynamic elements) as was discovered in the discussion for this PR: #90 .

Closes #93 

### How was it fixed?

Added this behavior and updated tests.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,7 +19,7 @@ common: &common
         when: on_fail
     - restore_cache:
         keys:
-          - v2-cache-{{ .Environment.CIRCLE_JOB }}-{{ checksum "setup.py" }}-{{ checksum "tox.ini" }}
+          - v4-cache-{{ .Environment.CIRCLE_JOB }}-{{ checksum "setup.py" }}-{{ checksum "tox.ini" }}
     - run:
         name: install dependencies
         command: pip install --user tox
@@ -33,7 +33,7 @@ common: &common
           - ~/.cache/pip
           - ~/.local
           - ./eggs
-        key: v3-cache-{{ .Environment.CIRCLE_JOB }}-{{ checksum "setup.py" }}-{{ checksum "tox.ini" }}
+        key: v4-cache-{{ .Environment.CIRCLE_JOB }}-{{ checksum "setup.py" }}-{{ checksum "tox.ini" }}
 
 jobs:
   doctest:
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -29,6 +29,7 @@ Table of Contents
     encoding
     decoding
     registry
+    nested_dynamic_arrays
     eth_abi
     releases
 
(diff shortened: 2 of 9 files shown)


carver (approved):
Looks good to me!

Have you run a gut check against a contract? It would be nice to make sure Solidity is interpreting the spec the same way we are.

For this manual check, I'm imagining a quick remix contract that returns a dynamic array value. Then check that the returned bytes are how we would have encoded them, as well. (and of course, that we can decode them)

davesque:
@carver Yes, that's a good idea.  Actually, the community member who originally identified this issue ( @tristan ) in pr #90 provided an example encoding which was being returned from a contract.  I did confirm that our implementation creates the same encoding.  However, it wouldn't hurt to check a more complicated use case.

pipermerriam (approved):

PR #95, opened 2018-08-22
Dynamic array support

### LINKED ISSUE #94: Backport dynamic array encoding to v1
### What was wrong?

#93 support is not available in a stable version (v1).

### How can it be fixed?

Backport #92 to the v1 branch

carver:
### What was wrong?

Closes #94 

### How was it fixed?

Backported #90 and #92 sequentially (plus some in-commit patches to make it v1-compatible)

#### Cute Animal Picture

Cute animal picture


carver:
@davesque Have you seen this ~"examples too slow" error from hypothesis on pypy before? I'm sure I can dig in and figure it out, just curious if you had a quick fix handy.

pipermerriam:
@carver probably worth just creating an issue and adding the `too_slow` to the ignored health checks.

carver:
Ok, ready for review, then.

davesque:
@carver My "quick fix" lately has just been to re-run the failed job until is passes :).  I believe the failing tests are the ones which do property testing to check coder class behavior.  They should actually be deleted since they aren't that useful.  That's the *real* fix.

davesque:
@carver Actually, I was wrong.  I think the issue was being caused by a recursive strategy that was generating too many leaves.  I made a one-line PR for that issue specifically.  I think you can revert the fix you added to this PR.

carver:
Cool, I removed the slow test exception. I'll wait to merge until you review, @davesque

davesque (approved):
I think this is good to go, Jason.

Issue #96, opened 2018-08-28
Figure out why some hypothesis tests are running too slow


carver:
### What was wrong?

See the test failure on https://github.com/ethereum/eth-abi/pull/95 and the subsequent patchup: https://github.com/ethereum/eth-abi/pull/95/commits/f0bf403639ceef5c09838037fe48a3029e81386d

### How can it be fixed?

:man_shrugging:

davesque:
Just made a PR for this with a fix that is probably more effective.  I'm pretty sure the issue was that there was a recursive hypothesis strategy which could generate up to 100 leaves on each level.

PR #97, opened 2018-08-28
Fix slow hypothesis tests

davesque:
### What was wrong?

See issue #96 .

### How was it fixed?

Restricted the number of leaves possible in data generated by recursive strategy.

#### Cute Animal Picture

Cute animal picture


davesque:
Not much here but figured I'd go through the typical review process anyway.

PR #99, opened 2018-09-17
Fix eth-typing major version constraint

carver:
### What was wrong?

The dependency on eth-typing permitted an upgrade to eth-typing v2

### How was it fixed?

The dependency was reduced from `<=2` to `<2`.

#### Cute Animal Picture

Cute animal picture


pipermerriam (approved):

davesque (approved):

davesque:
Whoops.  One little character makes a big difference.

PR #100, opened 2018-09-18
Fix eth-typing major version constraint

carver:
Backported from #99 master to v1

#### Cute Animal Picture

Cute animal picture


davesque (approved):

davesque:
Ha, that picture.

PR #102, opened 2018-10-14
Include value in validation failure message for AddressEncoder

### LINKED ISSUE #101: attempting to encode bytes of wrong size produces a misleading error message
* Version: 2.0.0b1
* Python: 3.7
* OS: macos

### What was wrong?

```
>>> eth_abi.encode_single("address", b"a"*20)
b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00aaaaaaaaaaaaaaaaaaaa'
```

```
>>> eth_abi.encode_single("address", b"a"*21)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/xuanji/gits/cf/plasma-cash/venv/lib/python3.7/site-packages/eth_abi/abi.py", line 43, in encode_single
    return encoder(arg)
  File "/Users/xuanji/gits/cf/plasma-cash/venv/lib/python3.7/site-packages/eth_abi/encoding.py", line 73, in __call__
    return self.encode(value)
  File "/Users/xuanji/gits/cf/plasma-cash/venv/lib/python3.7/site-packages/eth_abi/encoding.py", line 178, in encode
    self.validate_value(value)
  File "/Users/xuanji/gits/cf/plasma-cash/venv/lib/python3.7/site-packages/eth_abi/encoding.py", line 456, in validate_value
    cls.__name__,
eth_abi.exceptions.EncodingTypeError: Value of type <class 'bytes'> cannot be encoded by AddressEncoder
```

contrary to what the error message says, `bytes` can be encoded by AddressEncoder

### How can it be fixed?

stefanmendoza:
### What was wrong?

Fixes #101 

### How was it fixed?

It was suggested that we have a different message for `str` ([see here](https://github.com/ethereum/eth-abi/issues/101#issuecomment-429482386)), but that doesn't solve the issue described because the message still implies that the type is the issue when `bytes` is passed. Since `is_address` is just a boolean check and doesn't give context to what the failure was (i.e. malformed data vs. invalid type), I think the only solution here is to just include the value in the message and the consumer should at least have more context to deduce why the failure occurred. @pipermerriam, what are your thoughts? If these changes look good, I'm assuming we'd want a backport to 1.x?

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/eth_abi/encoding.py
+++ b/eth_abi/encoding.py
@@ -451,8 +451,8 @@ class AddressEncoder(Fixed32ByteSizeEncoder):
     def validate_value(cls, value):
         if not is_address(value):
             raise EncodingTypeError(
-                "Cannot encode value {0} of type {1} as an address using {2}".format(
-                    value,
+                "Cannot encode value {0} of type {1} using {2}".format(
+                    repr(value),
                     type(value),
                     cls.__name__,
                 )
--- a/eth_abi/encoding.py
+++ b/eth_abi/encoding.py
@@ -451,7 +451,8 @@ class AddressEncoder(Fixed32ByteSizeEncoder):
     def validate_value(cls, value):
         if not is_address(value):
             raise EncodingTypeError(
-                "Value of type {0} cannot be encoded by {1}".format(
+                "Cannot encode value {0} of type {1} as an address using {2}".format(
+                    value,
                     type(value),
                     cls.__name__,
                 )


pipermerriam (approved):

pipermerriam:
(I am still of the opinion that two different error messages might be better since:

1. bad values of either `bytes/string` indicate potentially malformed addresses (or just bad values)
2. non `bytes/string` types indicate a full type mismatch and a more egregious error.

But the current messaging does adequately cover both cases

pipermerriam:
I'll leave it to @davesque since he's been shepherding this library to make a final call and deal with issuing any patch releases next week.

stefanmendoza:
@pipermerriam sounds good! Thanks for looking at this though. 👍

@davesque - if you feel like this doesn't help provide more clarity, feel free to go ahead and close this.

eth_abi/encoding.py
```diff
@@ -451,7 +451,8 @@ class AddressEncoder(Fixed32ByteSizeEncoder):
     def validate_value(cls, value):
         if not is_address(value):
             raise EncodingTypeError(
-                "Value of type {0} cannot be encoded by {1}".format(
+                "Cannot encode value {0} of type {1} as an address using {2}".format(
+                    value,
```
davesque: Can we make this change?  Then I'll merge it.
```suggestion
                    repr(value),
```

davesque (changes requested):

eth_abi/encoding.py
```diff
@@ -451,7 +451,8 @@ class AddressEncoder(Fixed32ByteSizeEncoder):
     def validate_value(cls, value):
         if not is_address(value):
             raise EncodingTypeError(
-                "Value of type {0} cannot be encoded by {1}".format(
+                "Cannot encode value {0} of type {1} as an address using {2}".format(
```
davesque: Also, I'd prefer this wording instead.  I feel like the name of the encoder implies that an address is being encoded:
```suggestion
                "Cannot encode value {0} of type {1} using {2}".format(
```

davesque (changes requested):

eth_abi/encoding.py
stefanmendoza: Good call! Fixed:
https://github.com/ethereum/eth-abi/pull/102/commits/5ec58537473caf97f4c21dadab91dda8f4c8ce59

eth_abi/encoding.py
stefanmendoza: Fixed 👍 
https://github.com/ethereum/eth-abi/pull/102/commits/5ec58537473caf97f4c21dadab91dda8f4c8ce59

davesque (approved):

PR #110, opened 2018-12-02
updating to eth-typing to 2.0.0

javimaravillas:
### What was wrong?

eth-typing was fixed to the minor version of 2.0.0. Due to https://github.com/ethereum/py-evm/issues/1484, we need a new version with that dependency updated.

### How was it fixed?

Updating the dependency was enough. Test are passing. Didn't find any breaking change from 1.3 to 2.0.0, neither in code or documentation.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,7 +18,7 @@ setup(
     include_package_data=True,
     install_requires=[
         'eth-utils>=1.2.0,<2.0.0',
-        'eth-typing==2.0.0',
+        'eth-typing>=2.0.0,<3.0.0',
         'parsimonious>=0.8.0,<0.9.0',
     ],
     extras_require={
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -18,7 +18,7 @@
     include_package_data=True,
     install_requires=[
         'eth-utils>=1.2.0,<2.0.0',
-        'eth-typing<2',
+        'eth-typing==2.0.0',
```
davesque: We typically use version ranges for install requirements.  That makes it so that things are less likely to break when several packages are installed at once that all require `eth-typing`.
```suggestion
        'eth-typing>=2.0.0,<3.0.0',
```

davesque (changes requested):

PR #111, opened 2018-12-04
Update instructions for installing dev dependencies

lithp:
`requirements-dev.txt` does not exist as of
f89d0bea05e7ceb8eab7f6ab2a36152968dc7e56

### What was wrong?

The command for installing development dependencies referenced a non-existent file

### How was it fixed?

Instead of referencing the file, I'm referencing the correct section of `setup.py`

#### Cute Animal Picture

tiny-kitten

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,8 +21,10 @@ For documentation, visit [https://eth-abi.readthedocs.io/en/latest/](https://eth
 Clone the repository and then run:
 
 ```sh
-pip install -e . -r requirements-dev.txt
+pip install -e .[test]
 ```
+
+You might want to do this inside a virtualenv.
 
 ### Running the tests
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -21,9 +21,11 @@ For documentation, visit [https://eth-abi.readthedocs.io/en/latest/](https://eth
 Clone the repository and then run:
 
 ```sh
-pip install -e . -r requirements-dev.txt
+pip install -e .[test]
```
davesque:
```suggestion
pip install -e .[dev]
```

davesque (changes requested):

davesque:
Yeah, some of the READMEs/docs are inconsistent with some of these packages since they were created before we started using a project template (I think).  We should use the "dev" extras here since they include *all* extra packages and not just those required for running the tests.

README.md
lithp: Makes sense to me

PR #114, opened 2019-01-15
Registry copying functionality

davesque:
### What was wrong?

We needed a facility for copying existing registries so we can use eth-abi for web3.py's encodability checks.

### How was it fixed?

Added copying functionality to `ABIRegistry` and supporting classes.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/tests/test_registry/test_predicate_mapping.py
+++ b/tests/test_registry/test_predicate_mapping.py
@@ -1,4 +1,6 @@
+import copy
 import functools
+import itertools
 import operator
 
 import pytest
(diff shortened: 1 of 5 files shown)


tests/test_registry/test_predicate_mapping.py
```diff
@@ -18,39 +20,39 @@ def mapping():
 equals_bar = Equals('bar')
 
 
-def test_can_add_predicates_and_find_corresponding_values(mapping):
+def test_can_add_predicates_and_find_corresponding_values(mapping: PredicateMapping):
```
pipermerriam, line 23: We've historically not used type hinting in the tests.  Any reason this was added?

Reason being that we often want to violate type safety when testing invalid inputs and whatnot.

pipermerriam (approved):

carver:
Cool, nothing significant to add. I guess a new test could explicitly confirm what's in the docs: that a change doesn't get propagated from a copied registry to its source registry. (in addition to the `id()` checks)

PR #118, opened 2019-02-20
Add API methods to ABIType for web3

davesque:
### What was wrong?

Web3.py still uses the deprecated `process_type` function which assumes all type strings are terminal types (non-tuple types).

### How was it fixed?

This PR adds some API features to the AST classes whose instances are output by eth-abi's new parsing facilities.  These new API features can then be utilized in web3.py to remove the old `process_type` function.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/eth_abi/grammar.py
+++ b/eth_abi/grammar.py
@@ -185,6 +185,18 @@ class ABIType:
             ),
         )
 
+    @property
+    def is_array(self):
+        return self.arrlist is not None
+
+    @property
+    def is_dynamic(self):
+        raise NotImplementedError('Must implement `is_dynamic`')
+
+    @property
+    def _has_dynamic_arrlist(self):
+        return self.is_array and any(len(dim) == 0 for dim in self.arrlist)
+
 
 class TupleType(ABIType):
     """
@@ -229,6 +241,13 @@ class TupleType(ABIType):
         # of its components contain an invalid type such as "uint7"
         for c in self.components:
             c.validate()
+
+    @property
+    def is_dynamic(self):
+        if self._has_dynamic_arrlist:
+            return True
+
+        return any(c.is_dynamic for c in self.components)
 
 
 class BasicType(ABIType):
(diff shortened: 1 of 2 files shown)


pipermerriam (approved):
This looks solid.  I assume these are all internal APIs or is there a documentation update that needs to be made anywhere?

davesque:
@pipermerriam I suppose this part of eth-abi could maybe use a documentation update.  Since we're going to pull it into web3.py, it's starting to look like public API but I don't think there's any section in the docs for it.  I'd like to address that in a separate issue ticket if possible so it doesn't block the related web3 work for too long.

pipermerriam:
I'm fine with docs happening separately but they should happen asap so we don't forget.

tests/test_grammar.py
```diff
@@ -199,9 +199,109 @@ def test_normalize(type_str, normalized):
...
+    (
+        'bytes[2][2]',
+        'bytes',
+        'string',
+        '(int[],bool)',
+    ),
+)
+def test_abi_type_item_type(type_str):
+    abi_type = parse(type_str)
+    assert not abi_type.has_dynamic_arrlist
```
carver: This test looks repeated. Based on the name, maybe it was supposed to test something else?

eth_abi/grammar.py
```diff
@@ -185,6 +185,18 @@ def invalidate(self, error_msg):
...
+    @property
+    def is_array(self):
+        return self.arrlist is not None
+
+    @property
+    def is_dynamic(self):
+        raise NotImplementedError('Must implement `is_dynamic`')
+
+    @property
+    def has_dynamic_arrlist(self):
```
carver: Does `has_dynamic_arrlist` need to be public? Seems like it's mostly a helper property, so my instinct is to `_`-prefix it until we know we want to use it somewhere.

carver (approved):
LGTM. Pointing out one thing that maybe doesn't need to be public, which gives us more flexibility later if we want to refactor.

tests/test_grammar.py
davesque: Whoops!

PR #120, opened 2019-02-21
Change ABIType.__str__ to ABIType.to_type_str

davesque:
### What was wrong?

Piper [mentioned](https://github.com/ethereum/web3.py/pull/1249#discussion_r258565787) that the API of using the `str` builtin to convert `ABIType` instances to canonical type strings is a bit risky.

### How was it fixed?

Added a more explicit `to_canonical_type_str` function for this purpose.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/eth_abi/decoding.py
+++ b/eth_abi/decoding.py
@@ -174,7 +174,10 @@ class TupleDecoder(BaseDecoder):
 
     @parse_tuple_type_str
     def from_type_str(cls, abi_type, registry):
-        decoders = tuple(registry.get_decoder(str(c)) for c in abi_type.components)
+        decoders = tuple(
+            registry.get_decoder(c.to_type_str())
+            for c in abi_type.components
+        )
 
         return cls(decoders=decoders)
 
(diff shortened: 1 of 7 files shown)


davesque:
@pipermerriam By the way, I almost feel like I prefer the name `to_type_str` since it's more concise.  Thoughts?

pipermerriam:
I don't have a strong preference on `to_type_str` or `to_canonical_type_str`.  My *weak* inclincation towards having `canonical` in there was to serve a differentiator between things like `uint` vs `uint256` but that's a broader concept (mostly just in solidity source code and ?maybe? vyper source code) which doesn't seem to actually cross the boundaries into `eth-abi`.

So, probably just fine to leave out, but good to add *if* we find we have places where `eth-abi` is encountering non-caninical type strings.

pipermerriam (approved):

carver (approved):
Yeah, no strong preference. I think that in the context of an ABI-specific library, "Type String" only has to mean one thing. So `to_type_str()` seems acceptable. The ABI docs in Solidity do call it the canonical type string, or just "type" at times, so I don't have any better suggestions.

davesque:
@pipermerriam Actually, now that you mention it, the `int` type is parsable even though it's not canonical and the previous `__str__` implementation would just yield `int` back out.  That's by design and I feel that it should be kept that way since that's how the previous `process_type` and `collapse_type` functions worked.  So it actually shouldn't be called `to_canonical_type_str` at all.

PR #123, opened 2019-03-19
Rename testing modules to reflect new guidelines

davesque:
### What was wrong?

To reflect new guidelines in [this](https://github.com/ethereum/snake-charmers-tactical-manual/pull/40) tactical manual PR, renamed some testing modules.

#### Cute Animal Picture

Cute animal picture


carver (approved):

carver:
Only mildly related, maybe we should move `eth_abi.utils` -> `eth_abi._utils` which I think is a nice indication that they are intended as internal use modules (and we've been starting to move to that pattern in other repos)

kclowes:
Going through old issues. Thanks @davesque!

PR #124, opened 2019-03-19
Update officially supported python versions

davesque:
### What was wrong?

Looks like web3.py only tests against python 3.6 and 3.7.

### How was it fixed?

Updated testing config to do the same.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -48,30 +48,23 @@ jobs:
       - image: circleci/python:3.5
         environment:
           TOXENV: lint
-  py35-core:
-    <<: *common
-    docker:
-      - image: circleci/python:3.5
-        environment:
-          TOXENV: py35-core
   py36-core:
     <<: *common
     docker:
       - image: circleci/python:3.6
         environment:
           TOXENV: py36-core
-  pypy3-core:
+  py37-core:
     <<: *common
     docker:
-      - image: pypy
+      - image: circleci/python:3.7
         environment:
-          TOXENV: pypy3-core
+          TOXENV: py37-core
 workflows:
   version: 2
   test:
     jobs:
       - doctest
       - lint
-      - py35-core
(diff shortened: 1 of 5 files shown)


carver (commented):
Re: lint -- you could make isort fixups easy by adding the `lint-roll` make command from https://github.com/ethereum/web3.py/blob/a22c6cc898df6665814aafcfb797fc095201a470/Makefile#L27-L29

:+1: for dropping py3.5 in v2

:-1: for dropping pypy3 -- might as well keep support alive as long as possible in as many libraries as possible, until it becomes a huge burden

davesque:
Yep, I'll add that.

Wasn't there some talk about dropping pypy support?  I dunno.  I feel like it's a bit of a hassle to have to remember which packages do or don't support pypy.  Seems better to just decide one way or another that we support pypy.  Another problem I have with supporting pypy is that it often lags behind the current python sub version by about two versions.  That limits the possible use of nice, next gen language features.

davesque:
So dropping python3.5 support but keeping pypy3 support is really like doing nothing because pypy3 only implements 3.5 last I checked.

carver:
> I feel like it's a bit of a hassle to have to remember which packages do or don't support pypy. Seems better to just decide one way or another that we support pypy.

My logic is that it's much easier to notice that you're adding a pypy-unfriendly library in a failing PR than it is to go re-audit every library and every dependency of every library we maintain.

> So dropping python3.5 support but keeping pypy3 support is really like doing nothing because pypy3 only implements 3.5 last I checked.

Argh, okay, fine. I hate py3.5 slightly more than I hate having to go back and make every single library re-support pypy3 someday.

carver (approved):

davesque:
Hehe, yay! 🎉

PR #128, opened 2019-03-25
Add API for checking type validity against coder registrations

davesque:
### What was wrong?

The `is_encodable` method made some implicit assumptions that value encodability would only be checked for types recognizable by the registry.

### How was it fixed?

Added methods `is_encodable_type` and `has_encoder` to the public API and registry respectively to provide functionality for checking type validity.

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/eth_abi/exceptions.py
+++ b/eth_abi/exceptions.py
@@ -99,6 +99,11 @@ class NoEntriesFound(ValueError, PredicateMappingError):
     """
     Raised when no registration is found for a type string in a registry's
     internal mapping.
+
+    ..warning::
+
+        In a future version of ``eth-abi``, this error class will no longer
+        inherit from ``ValueError``.
     """
     pass
 
@@ -106,6 +111,13 @@ class NoEntriesFound(ValueError, PredicateMappingError):
 class MultipleEntriesFound(ValueError, PredicateMappingError):
     """
     Raised when multiple registrations are found for a type string in a
-    registry's internal mapping.
+    registry's internal mapping.  This error is non-recoverable and indicates
+    that a registry was configured incorrectly.  Registrations are expected to
+    cover completely distinct ranges of type strings.
+
+    ..warning::
+
+        In a future version of ``eth-abi``, this error class will no longer
+        inherit from ``ValueError``.
     """
     pass
--- a/eth_abi/exceptions.py
+++ b/eth_abi/exceptions.py
@@ -86,3 +86,26 @@ class ABITypeError(ValueError):
     that is not congruent with zero modulo eight).
     """
     pass
+
+
+class PredicateMappingError(Exception):
+    """
+    Raised when an error occurs with the registry's internal mapping.
+    """
+    pass
+
+
+class NoEntriesFound(ValueError, PredicateMappingError):
+    """
+    Raised when no registration is found for a type string in the registry's
+    internal mapping.
+    """
+    pass
+
+
+class MultipleEntriesFound(ValueError, PredicateMappingError):
+    """
+    Raised when multiple registrations are found for a type string in the
+    registry's internal mapping.
+    """
+    pass
--- a/eth_abi/exceptions.py
+++ b/eth_abi/exceptions.py
@@ -90,14 +90,14 @@ class ABITypeError(ValueError):
 
 class PredicateMappingError(Exception):
     """
-    Raised when an error occurs with the registry's internal mapping.
+    Raised when an error occurs in a registry's internal mapping.
     """
     pass
 
 
 class NoEntriesFound(ValueError, PredicateMappingError):
     """
-    Raised when no registration is found for a type string in the registry's
+    Raised when no registration is found for a type string in a registry's
     internal mapping.
     """
     pass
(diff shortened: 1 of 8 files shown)


eth_abi/exceptions.py
```diff
@@ -86,3 +86,26 @@ class ABITypeError(ValueError):
...
+class PredicateMappingError(Exception):
+    """
+    Raised when an error occurs in a registry's internal mapping.
+    """
+    pass
+
+
+class NoEntriesFound(ValueError, PredicateMappingError):
+    """
+    Raised when no registration is found for a type string in a registry's
```
pipermerriam, line 100: *maybe* leave a note inline with this docstring that the inheritance chain is scheduled to exclude `ValueError` in whatever major version bump you see us removing that at.

eth_abi/registry.py
```diff
@@ -434,6 +438,20 @@ def unregister(self, label: str) -> None:
     def get_encoder(self, type_str):
         return self._get_coder(self._encoders, type_str)
 
+    def has_encoder(self, type_str: abi.TypeStr) -> bool:
+        """
+        Returns ``True`` if an encoder is found for the given type string
+        ``type_str``.  Otherwise, returns ``False``.  Raises
+        :class:`~eth_abi.exceptions.MultipleEntriesFound` if multiple encoders
```
pipermerriam, line 449: This is sort of unfortunate but I see why it is like this and I can't really see a way around it since we allow arbitrary functions for the *predicates* which determine if something is encodable......

tests/test_registry/test_abi_registry.py
```diff
@@ -170,3 +167,31 @@ def test_copying_copies_internal_mappings(registry: ABIRegistry):
...
+    assert registry.has_encoder('address')
+
+    registry.register(
+        BaseEquals('address', with_sub=False),
+        encoding.AddressEncoder, decoding.AddressDecoder,
+        label='other_address',
+    )
+
+    with pytest.raises(exceptions.MultipleEntriesFound):
+        assert registry.has_encoder('address')
```
carver: Looks like all but one line in this test is a duplicate of `test_has_encoder_raises()`

carver (approved):
LGTM!

tests/test_registry/test_abi_registry.py
davesque: Ahh, yep.  Probably just copied and pasted wrong.

davesque:
@pipermerriam Any thoughts on this before I merge it?

pipermerriam (approved):
Evidently I never posted this review.  Oops.

eth_abi/registry.py
carver: Right, I think there is a distinction that is worth being really clear about in the exception docstrings:
- `NoEntriesFound` is *probably* a bug with the string type you are sending to eth-abi, unless you just forgot to add a type to the registry, but that should be apparent to the caller
- `MultipleEntriesFound` is a bug with a custom registry you built (can it be anything else?)

pipermerriam: Yeah, expanding docstrings for those might save us some support effort in github issues down the road.  `MultipleEntriesFound`

https://github.com/ethereum/eth-abi/blob/169600c01b3f991c4d2cc739d1b9492c254503a2/eth_abi/registry.py#L93-L97

Expanding that to include the codec values as well might help, as well as a more verbose error messages:

> Multiple codecs were found for the type string "something".  The matching codecs and predicates which matched were:  (<CodecClassA>, <function: is_codec_class_A>), (...).  This error happens when there are two codec classes registered for the same type string.  You may need to unregister one of these codecs or create a custom matching function for your codec.  See the documentation on 'registering custom codecs' for more information.

davesque: @carver @pipermerriam Yeah, that's the distinction, and I think the reason for this entire pull request essentially.  One exception is probably okay and easily recovered from.  The other indicates that the user of the library just did something wrong.  It's an expected property of a registry that the different registrations cover completely distinct regions of the input domain.

davesque: @pipermerriam At the point where that error is raised, there's no information available about what value is being encoded or decoded.  I'm also not so sure that information would really be very helpful since the registration matching happens completely independently of the particular value being coded.

carver: > @pipermerriam At the point where that error is raised, there's no information available about what value is being encoded or decoded. I'm also not so sure that information would really be very helpful since the registration matching happens completely independently of the particular value being coded.

Shouldn't the type string be available? I agree that the type string is the only thing relevant for either of these exceptions.

davesque: @carver The type string is already included in the error message.

carver: ~~I don't see where he was suggesting that the values go into the message.~~ _Nevermind I see it._ LGTM in current state :+1:

davesque: @carver Okay, actually I think he meant the *coders* associated with the registrations.  There's some confusion about terminology here since the registry tracks coders (i.e. encoders or decoders) and not codecs.

pipermerriam: Yeah, I was thinking about including the actual coder classes.  Current message looks fine.

PR #130, opened 2019-04-26
Add strategies to package release for generating test ABI values

davesque:
### What was wrong?

Some users have a need for hypothesis strategies similar to those used in eth-abi's test suite.  These strategies could be used to generate fake data for basic Ethereum ABI types.

### How was it fixed?

* Added `eth_abi.tools` submodule that contains new strategies to generate test data.
* Added "tools" extra requirements to be installed with `pip install eth-abi[tools]`.
* Added documentation for using data strategies.
* Added support functions for obtaining data generators for simple types.

#### Cute Animal Picture

Cute animal picture


davesque:
Left to do:
* Add support functions for getting data generators for simple types
* Add documentation

davesque:
Some concerns I have:
* Will exposing these hypothesis strategies make it harder to update strategies used by tests since they're part of the public API?
* The issue of duplicate dependencies in `setup.py` is weird and I don't like the solution I came up with.

davesque:
@pipermerriam @carver 

Would like your input on this.  Moving these testing strategies into the package release using the `<package>.tools` submodule pattern has created some interesting issues.  Namely, both the `tools` submodule *and* the test suite depend on hypothesis.  So it really seems proper to include the same hypothesis requirement in both the `tools` and `test` extras.  However, the tests *also* depend on the `tools` submodule.  So we really should include all of the `tools` extras in the `test` extras.

I put together a little hack [here](https://github.com/davesque/eth-abi/blob/efa46f485ef1ef0bba5580ded4257f7aefbcbeac/setup.py#L43-L61) which tries to resolve this, but it just feels weird.  Any thoughts on a better alternative?

carver:
Yeah, it feels a little hacky, but I think any solution here will. I have some suggestions that may be incremental improvements. Will make inline...

setup.py
```diff
@@ -6,11 +6,18 @@
 )
 
 extras_require = {
+    # Separate hypothesis extra to ensure same version across multiple extras
+    # that require it.
+    'hypothesis': [
```
carver: Yeah I don't love adding an extra just to be used later in the file. I think it's unintentionally committing us to a public API. (`eth-abi[hypothesis]`).

Maybe we're better off just separating `extras_require` into sequential definitions, like:
```py
extras_require = {}
extras_require['test'] = [
  "pytest...",
# etc
```

More ideas about executing that, below.

setup.py
```diff
@@ -32,12 +39,26 @@
...
 }
 
-extras_require['dev'] = (
-    extras_require['dev'] +
-    extras_require['test'] +
-    extras_require['lint'] +
-    extras_require['doc']
-)
+
+def include_extras(base, includes):
```
carver: Once we start doing sequential extra definitions, we can pass in the dependencies explicitly here (rather than force them to be named ahead of time in the extras).

I'm thinking an API like `include_extras(base_list, *supplemental_lists)`:
```py
hypothesis_requirement = "hypothesis>=3.6.1"

extras_require['tools'] = [hypothesis_requirement]
extras_require['test'] = [
    "pytest...",
    ...,
    hypothesis_requirement,
]

extras_require['dev'] = include_extras(
  [
    "bumpversion...",
    ...,
  ],
  extras_require['test'],
  extras_require['lint'],
  extras_require['doc'],
)
```

carver (commented):
I think we can use some more direct manipulation instead of `include_extras()` in a couple places.

setup.py
davesque: Yeah, I like that better.  I think it was the mutable, closure-centered nature of how I was doing it that bothered me.

carver:
If we find that we like this after a while and/or it comes up in at least one more repo, then we can port the setup.py changes back to the project template.

pipermerriam (commented):
Looking at the `strategies.py` file there are a lot of things in there that are probably not intended to be made public.  Various functional utils that are used to manipulate the strategies.  I think it might be good to do a little code organization work to separate the stuff that's intended to be public from the stuff that isn't.

davesque:
@pipermerriam Yep, agreed.  Work is still ongoing.

davesque:
@pipermerriam Actually, since you pointed that out, I'm realizing that possibly very few of the testing strategies would be useful for the use case that @fubuloubu had in mind.  I'm going to take an alternate approach of making some strategies that I think would suit his needs and then see which ones overlap with the testing strategies we already have.  Thanks for making me take a closer look at that.

davesque:
And as a reminder to everyone, the PR description states the changes that need to be made before the PR would be considered complete, but most of those changes still haven't been added.

fubuloubu:
Yeah, that's a good takeaway. Really, I would only see value in generating ABI types to use as function inputs for test cases, not for the reverse (creating raw byte strings that parse to ABI types)

carver:
> And as a reminder to everyone, the PR description states the changes that need to be made before the PR would be considered complete, but most of those changes still haven't been added.

I recommend adding the `[ ]` after the bullets to turn them into checkboxes. (and check off any that are complete) Then the progress will be very apparent to reviewers.

fubuloubu (approved):
This is seriously awesome!

fubuloubu:
Verified the API works very well for the intended use case: https://github.com/fubuloubu/hypothesis-ethereum

eth_abi/utils/numeric.py
```diff
@@ -51,3 +54,24 @@ def compute_signed_fixed_bounds(
...
     return lower, upper
+
+
+def scale_places(places: int) -> Callable[[decimal.Decimal], decimal.Decimal]:
+    """
+    Returns a function that shifts the decimal point of decimal values to the
+    right by ``places`` places.
+    """
+    with decimal.localcontext(abi_decimal_context):
+        scaling_factor = TEN ** int(-places)
```
carver: Maybe instead of casting, this should validate `places` and raise an exception if it's not already an `int`. (Or is there some legitimate case to use `places=0.5`?)

setup.py
```diff
@@ -5,38 +5,53 @@
...
-        "flake8==3.4.1",
-        "isort>=4.2.15,<5",
-        "mypy==0.620",
-    ],
-    'doc': [
-        "Sphinx>=1.6.5,<2",
-        "sphinx_rtd_theme>=0.1.9",
-    ],
-    'dev': [
+    extras_require['tools'],
```
carver, line 33: I'm not sure `tools` needs to be merged into `test`. We can add `tools` as an explicit extra in the jobs that need it, and hypothesis is already explicitly added.

setup.py
```diff
@@ -5,38 +5,53 @@
...
         "ipython",
     ],
-}
-
-extras_require['dev'] = (
-    extras_require['dev'] +
-    extras_require['test'] +
-    extras_require['lint'] +
-    extras_require['doc']
+    extras_require['test'],
```
carver, line 52: If you agree with the previous comment, maybe `tools` would be added here, in case things beyond hypothesis get added. NBD either way.

tests/test_tools.py
```diff
@@ -0,0 +1,72 @@
...
+    (
+        ('uint8', 'integers(min_value=0, max_value=255)'),
+        ('uint16', 'integers(min_value=0, max_value=65535)'),
+        ('uint', 'integers(min_value=0, max_value=115792089237316195423570985008687907853269984665640564039457584007913129639935)'),  # noqa: E501
+        ('uint256', 'integers(min_value=0, max_value=115792089237316195423570985008687907853269984665640564039457584007913129639935)'),  # noqa: E501
+        ('int8', 'integers(min_value=-128, max_value=127)'),
+        ('int16', 'integers(min_value=-32768, max_value=32767)'),
+        ('int', 'integers(min_value=-57896044618658097711785492504343953926634992332820282019728792003956564819968, max_value=57896044618658097711785492504343953926634992332820282019728792003956564819967)'),  # noqa: E501
+        ('int256', 'integers(min_value=-57896044618658097711785492504343953926634992332820282019728792003956564819968, max_value=57896044618658097711785492504343953926634992332820282019728792003956564819967)'),  # noqa: E501
+        ('address', 'binary(min_size=20, max_size=20).map(encode_hex)'),
```
carver: Could map to `to_checksum_address`, for better web3 compatibility.

carver (approved):
Nice

eth_abi/utils/numeric.py
davesque: Yeah, that's a better approach.

setup.py
davesque: Would that make it harder for contributors to run tests locally without using tox i.e. by just doing `pytest`?  The job extras only get added by tox.  If the `tools` submodule adds requirements other than hypothesis, then you wouldn't be able to just do a vanilla `pytest` run without getting import errors.

davesque: I guess I have an expectation that, if I do `pip install -e .[test]`, then I've got everything I need to run the tests with `pytest`.  I also expect that the test suite covers everything, including optional submodules.

carver: > Would that make it harder for contributors to run tests locally without using tox i.e. by just doing `pytest`? The job extras only get added by tox. If the `tools` submodule adds requirements other than hypothesis, then you wouldn't be able to just do a vanilla `pytest` run without getting import errors.

I think this scenario is what `[dev]` is for. Sometimes extras are even mutually exclusive, like `eth-hash[pycryptodome]` and `eth-hash[pysha3]`, so there's no way to totally abstract away the things that tox is doing. My approach is that `[test]` is for adding the tools you need to run tests, not the things *being tested*. So if you want to test some extra that is not installed by default, then tox (or you, if outside of tox) will need to know to install it explicitly, or let `[dev]` install it for you.

In the end, this is all very academic, since `hypothesis` is installed by tests anyway. So I think we can kick the can down the road until tools has something that's not in `tests`.

davesque: > My approach is that [test] is for adding the tools you need to run tests, not the things being tested.

Gotcha, makes sense.

> So I think we can kick the can down the road until tools has something that's not in tests.

👍

Issue #135, opened 2019-11-13
Modelling of trailing zeros in ContextFramesBytesIO


JoranHonig:
* Version: 2.0.0

### What was wrong?
The eth-abi decoder models the calldata of a transaction as a finite length datastream.
However, the evm models calldata as an infinite length bytestream. 
Returning zero bytes when the transaction does not specify the bytes requested.

The tests in this repository do seem to indicate that the decoder is intentionally modelled this way.
I'm not sure why this is the case. It'd be cool if there is an additional option to model the bytestream as infinite.

### How can it be fixed?

I have a local branch where I adapted the `ContextFramesBytesIO` object to mimic the behaviour of calldata in the evm. 
I'm happy to submit a pr if this is something that could be added to the repository.

JoranHonig:
For reference: https://github.com/JoranHonig/eth-abi/blob/master/eth_abi/decoding.py#L111
This link points to function that alters the behaviour.
 I'm not very familiar with the codebase, so there might be some edgecases that I haven't thought of 😅

davesque:
Hey @JoranHonig , thanks for getting in touch.

Can you give an example of where this particular issue is causing problems for you?  I know there's a conceptual difference between the way the EVM specifies the behavior of call data (as a virtual infinite byte string) and how eth-abi views the data its sending over the wire (as finite byte strings).  But I'm having a hard time imagining how this would cause problems.

JoranHonig:
Hi @davesque, thanks for your response!

I'm trying to decode automatically generated testcases (in the shape of a series of transactions).
The automatic generation of these testcases is done by Mythril (on the bytecode). During it's analysis, Mythril will try to generate transaction traces for the different issues it uncovers. However, because we analyze the bytecode we have this issue that calldata is infinite.
Because we don't know the abi we don't really know how much trailing zero's are needed for the decoder. This is why we chose to remove any trailing zero's if possible (/ keep the calldatasize as low as possible). In the end, mythril will give transaction traces that would work on the evm, but are not decodable using eth-abi.

Long story, hope it makes sense 😅 

I'm also wondering if a similar issue could occur outside of mythril. Wouldn't wallets also benefit from removing trailing zeros? That should reduce gas cost, while not having any negative effects on the execution (since it does not matter from the evm's perspective).
These transactions would also not be decodable right?

davesque:
[...]

PR #136, opened 2019-11-26
Add ABIDecoder.stream_class property

davesque:
### What was wrong?

A user had the need to customize the behavior of the `ContextFramesBytesIO` class as was discussed in #135 .

### How was it fixed?

Added a class property called `stream_class` to `ABIDecoder` which can be overridden in subclasses.

#### Cute Animal Picture

Cute animal picture


davesque:
I don't think we really need to update the documentation or tests for this.

cburgdorf (approved):
LGTM :+1:

PR #137, opened 2019-12-10
Add alias "byte" for "bytes1"

davesque:
### What was wrong?

Just realized recently that the type "byte" is an official alias "bytes1" according to the Solidity docs here: https://solidity.readthedocs.io/en/v0.5.13/types.html#fixed-size-byte-arrays

### How was it fixed?

Modified `normalize` function to replace instance of "byte" with "bytes1".

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/eth_abi/grammar.py
+++ b/eth_abi/grammar.py
@@ -405,6 +405,7 @@ TYPE_ALIASES = {
     'fixed': 'fixed128x18',
     'ufixed': 'ufixed128x18',
     'function': 'bytes24',
+    'byte': 'bytes1',
 }
 
 TYPE_ALIAS_RE = re.compile(r'\b({})\b'.format(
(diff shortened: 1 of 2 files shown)


davesque:
I'll probably cut a patch release (v2.0.1) after this gets merged.

cburgdorf (approved):
LGTM

carver (approved):
> I'll probably cut a patch release (v2.0.1) after this gets merged.

Choosing `patch` is because you are treating it as a bug that we didn't already handle the `byte` type? I might have picked minor, but patch is totally reasonable, too. I just double-checked in the docs that `byte` has been around since at least Solidity v0.3, so it's not a new feature in Solidity ABIs.

davesque:
@carver Yeah, I was thinking:
1. It's backwards compatible
2. It's minor
3. It's kind of a bug

davesque:
@carver Actually, I'll cut a minor since I just remembered another feature was added: custom stream class support on the `ABIDecoder` class.

PR #139, opened 2020-01-24
use self reference rather than super

JoranHonig:
This change aligns the use of seek with the use in pop_frame
Additionally, inheritants can now override the behaviour of the seek call in seek_in_frame(...)

### What was wrong?
The call to seek in seek_in_frame() was using super() to get a reference to the function seek().

### How was it fixed?
super() -> self

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/eth_abi/decoding.py
+++ b/eth_abi/decoding.py
@@ -81,7 +81,7 @@ class ContextFramesBytesIO(io.BytesIO):
         """
         Seeks relative to the total offset of the current contextual frames.
         """
-        super().seek(self._total_offset + pos, *args, **kwargs)
+        self.seek(self._total_offset + pos, *args, **kwargs)
 
     def push_frame(self, offset):
         """


JoranHonig:
@the reviewer of this pr
I thought about opening an issue for this change, but it was easier to just submit a PR right away.

Let me know if you would like a larger ratification for the change!

davesque (approved):
LGTM.  I think this was a little oversight on my part.

carver:
> @the reviewer of this pr

Poor Thomas Eichmann probably gets this a lot :laughing: 

> I thought about opening an issue for this change, but it was easier to just submit a PR right away.

Thanks for the PR!

> LGTM. I think this was a little oversight on my part.

Yup, LGTM, too. Thanks for taking a look @davesque

carver (approved):

carver:
@JoranHonig sorry it took a while to get this out. Thanks for the contribution! It's available in v2.1.1

Issue #147, opened 2021-03-19
`is_encodable_type` does not function properly with mispelled integer types


fubuloubu:
## What was wrong?

### Code that produced the error

```py
from eth_abi import is_encodable_type
is_encodable_type("uint25") 
```

### Full error output 

```sh
ABITypeError: For 'uint25' type at column 1 in 'uint25': integer size must be multiple of 8 
```

### Expected Result

_This section may be deleted if the expectation is "don't crash"._

```sh
False
``` 

### Environment

```sh
Python version:                                                                                                                                                                                             
3.8.7 (default, Jan 24 2021, 11:57:42)                                                                                                                                                                      
[GCC 9.3.0]                                                                                                                                                                                                 
                                                                                                                                                                                                            
Operating System: Linux-5.6.0-1048-oem-x86_64-with-glibc2.29                                                                                                                                                
                                                                                                                                                                                                            
pip freeze result:                                                                                                                                                                                          
alabaster==0.7.12                                                                                                                                                                                           
apipkg==1.5                                                                                                                                                                                                 
appdirs==1.4.4                                                                                                                                                                                              
atomicwrites==1.4.0                                                                                                                                                                                         
attrs==20.3.0                                                                                                                                                                                               
Babel==2.9.0                                                                                                                                                                                                
backcall==0.2.0                                                                                                                                                                                             
bitarray==1.2.2                                                                                                                                                                                             
bleach==3.3.0                                                                                                                                                                                               
certifi==2020.12.5                                                                                                                                                                                          
cffi==1.14.5                                                                                                                                                                                                
chardet==4.0.0                                                                                                                                                                                              
cryptography==3.4.6                                                                                                                                                                                         
cytoolz==0.11.0                                                                                                                                                                                             
dataclassy==0.7.2                                                                                                                                                                                           
decorator==4.4.2                                                                                                                                                                                            
distlib==0.3.1                                                                                                                                                                                              
docutils==0.16                                                                                                                                                                                              
eth-abi==2.1.1
eth-hash==0.3.1
eth-keyfile==0.5.1
eth-keys==0.3.3
eth-rlp==0.2.1
eth-typing==2.2.2
eth-utils==1.10.0
execnet==1.8.0
filelock==3.0.12
hexbytes==0.2.1
hypothesis==4.57.1
idna==2.10
imagesize==1.2.0
importlib-metadata==3.7.3
ipdb==0.13.7
ipython==7.21.0
ipython-genutils==0.2.0
[...]

PR #148, opened 2021-03-19
fix: catch ABITypeError exceptions when checking type

fubuloubu:
## What was wrong?

Issue #147

## How was it fixed?

Catch `ABITypeError` and return False


kclowes (approved):
Sorry it's been so long. Thanks for the PR!

PR #155, opened 2022-01-07
Update mypy and add python 3.8 support

kclowes:
## What was wrong?
Going through and updating support for libraries. Old mypy doesn't work with python 3.8 (`fails with error: AttributeError: 'FlakesChecker' object has no attribute 'CONSTANT'`).

## How was it fixed?
Updated mypy to the latest version and added tests for python 3.8. Also had to update flake8 as well. 

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,13 +39,13 @@ jobs:
   docs:
     <<: *common
     docker:
-      - image: circleci/python:3.6
+      - image: circleci/python:3.8
         environment:
           TOXENV: docs
   lint:
     <<: *common
     docker:
-      - image: circleci/python:3.6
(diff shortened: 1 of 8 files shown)


fselmo (approved):
LGTM. Should we go ahead and remove python 3.6 support as well? 🤔

_edit: Also curious if 3.9 build would work as-is but otherwise we can add that in another PR_

kclowes:
I'll add 3.9 in another PR. I've been hesitant to drop 3.6, because I'm afraid it might kick off another dependency hell situation, but might as well do that while we're at it. I'll do that in another PR too. Thanks!

PR #156, opened 2022-01-12
Drop python 3.6 support, add python 3.9 and 3.10 support

kclowes:
## What was wrong?
Python 3.6 is no longer officially supported, so dropped that. Added support for more modern pythons - 3.9 & 3.10. 

## How was it fixed?
Added CI runs for 3.9, and 3.10. Removed CI runs for 3.6. 

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,13 +39,13 @@ jobs:
   docs:
     <<: *common
     docker:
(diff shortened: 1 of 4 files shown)


fselmo (approved):
lgtm! 🚢 💨

PR #158, opened 2022-02-10
Investigate and resolve discrepancies with solidity's abi.encode()

### LINKED ISSUE #151: Padding of empty `bytes`
What is the reason for padding with 32 zeros here for an empty bytes array?
https://github.com/ethereum/eth-abi/blob/decaadcdf7a33020846aafd9ba36859c1ab82909/eth_abi/encoding.py#L535

I think it's different from what happens here, namely no padding
https://github.com/ethereumjs/ethereumjs-abi/blob/ee3994657fa7a427238e6ba92a84d0b529bbcde0/lib/index.js#L146

I know the latter is deprecated but it seems this is how it is done by the opensea.io frontend. 

Are both correct but according to different standards/versions?

Thank you.

### LINKED ISSUE #157: `encode_abi()` does not yield the same result as Solidity's `abi.encode()`
## What was wrong?

- `eth-abi`'s `encode_abi()` does not match exactly what Solidity's abi `encode()` method produces. It likely has to do with `0`-padding among possibly other things. It isn't broken in the sense that the instructions for the EVM end up being the same (same bytes are sent just represented differently)... but this leads to issues with hashes not matching and, also, using more `gas` - and thus a less efficient transaction.

- See conversation in ethereum/web3.py#1065
- Possibly related to ethereum/web3.py#1812

### Expected Result

- The ideal scenario is that `eth-abi` should match exactly what is produced via Solidity's `encode()` method.

## How can it be fixed?

- Investigate where extraneous padding with zero's may occur and look into removing this extra bit of padding
- Investigate if there is anything else going on that is producing a different representation of the encoded data

fselmo:
## What was wrong?

Related issues: 
- closes #151
- closes #157

## How was it fixed?

It seems that, for `encode_abi()`, extra 32-byte zero padding was happening in a few places, compared to what Solidity's `abi.encode()` returns. This seems to be the case when dealing with empty dynamic types, including empty dynamic arrays.

i.e. 

- `encode_abi(['string'], [''])` vs Solidity `abi.encode('')`
- `encode_abi(['string[]'], [[]])` vs Solidity `string[] sl; abi.encode(sl);`

These discrepancies still yielded correct instructions for the EVM but ultimately led to different hashes and higher gas values for transactions encoded this way.

### To-Do

- [x] Clean up commit history
- [x] Add cute animal picture

### DIFF
--- a/eth_abi/encoding.py
+++ b/eth_abi/encoding.py
@@ -159,7 +159,6 @@ class TupleEncoder(BaseEncoder):
         )
 
         encoded_value = b''.join(head_chunks + tuple(tail_chunks))
-
         return encoded_value
 
     @parse_tuple_type_str
@@ -530,16 +529,12 @@ class ByteStringEncoder(BaseEncoder):
     @classmethod
     def encode(cls, value):
         cls.validate_value(value)
+        value_length = len(value)
 
-        if not value:
-            padded_value = b'\x00' * 32
-        else:
-            padded_value = zpad_right(value, ceil32(len(value)))
+        encoded_size = encode_uint_256(value_length)
+        padded_value = zpad_right(value, ceil32(value_length))
 
-        encoded_size = encode_uint_256(len(value))
-        encoded_value = encoded_size + padded_value
-
-        return encoded_value
+        return encoded_size + padded_value
 
     @parse_type_str('bytes')
     def from_type_str(cls, abi_type, registry):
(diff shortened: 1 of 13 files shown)


pacrob (approved):
lgtm!

kclowes (approved):
🎉 Thanks for digging in here!

MatthiasLohr:
Thx!

Any idea when this gets officially released (and even used by web3py)?

kclowes:
We'll probably release `eth-abi` next week sometime, and this will probably make it into web3 v6-beta.2 or beta.3.

MatthiasLohr:
No chance of a patch release for web3py 5.28.*? Really struggling issue right now...

kclowes:
I'm hesitant to pull it into web3py 5.28 because this is technically a breaking change (though arguably it's a bugfix which is a little less black and white). Most of the breaking changes in web3 v6 right now are dependency-related, so I _hope_ it won't be too bad to update. The other big breaking change is dropping support for python 3.6. Any chance you can try updating to web3 v6 and let me know how it goes? If you run into issues, let me know and I'll consider releasing a patch.

yixun-alpha:
Hi, ran into this again on `web3py` v6 beta (on `v3.0.1` of eth-abi); wondering if the above can be merged into `v3` of eth-abi? thanks!

PR #171, opened 2022-05-19
docs: correct decode_abi method name

antazoey:
## What was wrong?

The example in the doc was using a different (perhaps older?) name for the `decode_abi` method.

## How was it fixed?

use known import

### To-Do

- [ ] Clean up commit history

- [ ] Add entry to the [release notes](https://github.com/ethereum/eth-abi/blob/master/newsfragments/README.md)

#### Cute Animal Picture


kclowes:
Thanks for the PR. I suspect you're actually looking at the `latest` version of the docs that hasn't been released yet. I'll see if I can make `stable` the default on the read the docs app

pacrob:
`encode_abi` and `decode_abi` were renamed to `encode` and `decode` in PR #161.

PR #172, opened 2022-05-19
Add make notes command, exclude venv in MANIFEST and gitignore

kclowes:
## What was wrong?
I noticed a few things when I was trying to release v2.x.
- There was no `make notes` command to compile release notes
- venv files weren't being gitignored
- venv files had the potential to make their way into the dist

## How was it fixed?
- Added a few commands to the Makefile.
- added venvs to .gitignore
- added venv files to be excluded in the MANIFEST.in

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,7 +50,12 @@ docs: build-docs
 linux-docs: build-docs
 	xdg-open docs/_build/html/index.html
 
-notes:
+check-bump:
+ifndef bump
+	$(error bump must be set, typically: major, minor, patch, or devnum)
+endif
+
+notes: check-bump
 	# Let UPCOMING_VERSION be the version that is used for the current bump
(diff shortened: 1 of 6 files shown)


pacrob (approved):
sorry for delay, lgtm!

Makefile
```diff
@@ -24,14 +24,14 @@ clean-pyc:
 	find . -name '*~' -exec rm -f {} +
 
 lint:
-	tox -e lint
+	tox -elint
```
fselmo: Just curious, was there a reason for this change? It seems it still works as `tox -elint` but the usage docs examples separate the environment from the `-e`.

fselmo (approved):
Just a nit / question on the tox env command change. Also we talked about updating the `Makefile` to match the template project repo since this it has been updated in the time this PR was open (sorry this took a while!).

lgtm though 👍

Makefile
kclowes: Oops, nope. Not sure where this change came from.

PR #177, opened 2022-07-13
Add black

kclowes:
## What was wrong?

Adds black linting to eth-abi.

## How was it fixed?

Added black linting, had to upgrade pydocstyle because it was conflicting with black. See https://github.com/psf/black/issues/709. This is technically a breaking change with the pydocstyle upgrade, but I was fighting with black when I was trying to release web3, so I think we should put this in v3 instead of just putting it in master. But open to hearing opinions.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture


fselmo (approved):
lgtm... one more down! 🎉 ... good to upgrade `pydocstyle` too 👍

PR #180, opened 2022-07-14
Ignore black revs

kclowes:
## What was wrong?
Git blame gets clobbered after adding black linting to a codebase. This will ignore that SHA in git blame. 

## How was it fixed?

Added `.git-blame-ignore-revs` file and the SHA that added `black`

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture


kclowes:
Oooookay, this is branched off of the right thing now. Thanks reviewers!

fselmo (approved):
yessss 👍

PR #181, opened 2022-07-14
Update misc release things

kclowes:
## What was wrong?
Added updates to the stable/2.x branch, and needed to pull them in here. 

## How was it fixed?

cherry-picked updates from #172 and applied them to stable/3.x

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -3,14 +3,14 @@ current_version = 3.0.0
 commit = True
 tag = True
 parse = (?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)(-(?P<stage>[^.]*)\.(?P<devnum>\d+))?
-serialize = 
+serialize =
 	{major}.{minor}.{patch}-{stage}.{devnum}
 	{major}.{minor}.{patch}
 
 [bumpversion:part:stage]
 optional_value = stable
 first_value = stable
-values = 
+values =
 	alpha
 	beta
 	stable
@@ -18,5 +18,5 @@ values =
 [bumpversion:part:devnum]
 
 [bumpversion:file:setup.py]
-search = version='{current_version}',
-replace = version='{new_version}',
+search = version="{current_version}",
+replace = version="{new_version}",
--- a/Makefile
+++ b/Makefile
@@ -50,7 +50,12 @@ docs: build-docs
 linux-docs: build-docs
 	xdg-open docs/_build/html/index.html
 
-notes:
+check-bump:
+ifndef bump
+	$(error bump must be set, typically: major, minor, patch, or devnum)
+endif
+
+notes: check-bump
 	# Let UPCOMING_VERSION be the version that is used for the current bump
 	$(eval UPCOMING_VERSION=$(shell bumpversion $(bump) --dry-run --list | grep new_version= | sed 's/new_version=//g'))
 	# Now generate the release notes to have them included in the release commit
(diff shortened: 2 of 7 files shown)


fselmo (changes requested):
I noticed a few small things and then one bigger breaking thing caused by running `black` on some files.

_Small things and some nits:_

Since we're updating the `Makefile` I compared it with the python project template `Makefile` and noticed we could add / change a few things:

- `find . -name '__pycache__' -exec rm -rf {} +` at the end of the `clean-pyc` command
- We could split out the `validate-docs` command from there if we wanted to have that separated
- `make notes` is missing "build" from the towncrier line, it should be: `towncrier build --yes --version $(UPCOMING_VERSION)`
- nit cleanup (since both seem to work): `tox elint` -> `tox e lint` in `make lint` command

I also noticed some warnings when running `make notes` related to titles in the docstrings [here](https://github.com/ethereum/eth-abi/blob/master/eth_abi/exceptions.py#L27) and [here](https://github.com/ethereum/eth-abi/blob/master/eth_abi/exceptions.py#L45) that can be silenced by just removing the `-----------` which makes it a title. `black` also seems to be happy with that so this works. Unless we want to keep that formatting in which case adding one more `-` since it is too short reduces the noise from the warning to just an "unexpected title" and it isn't as loud of a warning (not bright red). I'd be good with removing them altogether or extending.

---
_Big thing caused by `black` commit:_

I noticed that `bumpversion` can't find the library version anymore because it annoyingly only looks for single quotes in the line `version='3.0.0'` (in `setup.py`) and we (maybe accidentally?) ran `black` on `setup.py` which changed it to double quotes. We should change it back to single quotes as `version='3.0.0'` and then it does find the version correctly.

We also ran `black` on `docs/conf.py` which changed [this line](https://github.com/ethereum/eth-abi/blob/master/docs/conf.py#L26-L26) which we will need to change back to `setup_version = line.split("'")[1]` once the above `version` change is made.

Unless there's another way to get around it, but currently those are the changes I made locally that got it all building.

kclowes:
Thank you for the thorough review! Pydocstyle was erroring out without the `-------`, which is why it got added in the first place. I'll see what I can do about silencing those errors in either `pydocstyle` or `make notes`. I'll also fix the version thing from the black changes. I thought I changed it, but maybe linters are fighting with one another.

.bumpversion.cfg
```diff
@@ -3,20 +3,20 @@ current_version = 3.0.0
...
 	alpha
 	beta
 	stable
 
 [bumpversion:part:devnum]
 
 [bumpversion:file:setup.py]
-search = version='{current_version}',
-replace = version='{new_version}',
+search = version="{current_version}",
```
fselmo, line 21: Ah, nice I like this answer much better than removing the `black` formatting 👌

kclowes:
@fselmo this is ready for another look when you get a chance! ty!

Makefile
```diff
@@ -41,25 +42,33 @@ build-docs:
 	$(MAKE) -C docs clean
 	$(MAKE) -C docs html
 	$(MAKE) -C docs doctest
+
+validate-docs: build-docs
 	./newsfragments/validate_files.py
-	towncrier --draft --version preview
+	towncrier build --draft --version preview
```
fselmo, line 48: I guess I didn't realize this was a `build` command as well... I'm noticing as I look across the libraries that I don't think it makes a difference but maybe at some point using `build` vs not using `build` might end up with different behaviors. Something to keep in mind I suppose.

fselmo (approved):
lgtm! :shipit:

Makefile
kclowes: Without the `build`, it looks like it just returns the towncrier version. I'll keep an eye out and update when I come across those elsewhere!

PR #182, opened 2022-07-14
Add black check instead of running black

kclowes:
## What was wrong?

forgot to add the `--check` flag to the black command on CI.

## How was it fixed?

Added it!

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture


fselmo (approved):
Ah yes, nice catch 👍 ... lint CI job does read "**would be** left unchanged" now 👀 ... lgtm

PR #196, opened 2022-11-21
Remove venv files from package

kclowes:
## What was wrong?

the `exclude` line in the MANIFEST.in wasn't doing the right thing. 

## How was it fixed?

Added the correct command to prune virtualenv files from the dist.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture


fselmo (approved):
lgtm 👍🏼

PR #204, opened 2023-04-04
Fix requirements docs txt

pacrob:
## What was wrong?

readthedocs was building with the wrong versions, related to https://github.com/ethereum/ethereum-python-project-template/issues/40

## How was it fixed?

Discovered a `.readthedocs.yml` format that renders `requirements-docs.txt` unnecessary.

Removed `requirements-docs.txt`, moved it all to `.readthedocs.yml`.

Also, prefer `.yml` to `.yaml`.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

image


newsfragments/204.internal.rst
```diff
@@ -0,0 +1 @@
+Moved requirements-docs.txt info info .readthedocs.yml
```
fselmo:
```suggestion
Moved requirements-docs.txt info into .readthedocs.yml
```

.readthedocs.yml
```diff
@@ -0,0 +1,9 @@
+version: 2
```
fselmo, line 1: This looks like it differs from the template repo... what's the end result for this setup across all projects? We should maybe start there.

edit: I imagine it's this setup... and that lgtm... we should add it to the template repo if so.

fselmo (approved):
Ah, I missed the comment links... I see the context. This lgtm as the tester 👌🏼 

So, just the typo in the newsfragment :)

newsfragments/204.internal.rst
pacrob: thanks for catching!

PR #212, opened 2023-08-17
Disallow zero sized types

fselmo:
### What was wrong?

- Solidity doesn't support empty structs / tuples; we should not support them in `eth-abi`.

### How was it fixed?

- Validate against zero-sized tuples and provide a good message for these errors.
- Add tests and update current tests for empty tuples.

### Todo:
- [x] Clean up commit history

#### Cute Animal Picture

The Sunda Colugo

### DIFF
--- a/eth_abi/codec.py
+++ b/eth_abi/codec.py
@@ -24,6 +24,7 @@ from eth_abi.registry import (
 )
 from eth_abi.utils.validation import (
     validate_bytes_param,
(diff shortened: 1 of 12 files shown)


reedsa (approved):
thanks for the walkthrough, lgtm!

reedsa (approved):
Looks great, thanks for taking the extra time on this one.

PR #217, opened 2023-12-09
Upgrade template

### LINKED ISSUE #207: typing/linting tools review
`eth-abi` is ignoring a bunch of typing and linting rules. Review for updating the code so we don't need to limit what we type/lint.

all in `tox.ini` -> [common-lint]:
 - [ ] `mypy` uses `--follow-imports=silent -p eth_abi.utils`, can it just be `-p eth_abi`?
 - [ ] `flake8` uses `--extend-ignore=B019,B011,W503,E203`, can we remove `--extend-ignore` and reserve only `E203` in the `flake8` section? (required to play nice with `black`
 - [ ] `pydocstyle` uses `--add-ignore=D415`, can we remove it?

pacrob:
### What was wrong?

Merge changes from the project template, notably:
 - drop python 3.7
 - use `pre-commit` for linting
 - change the name of the `master` branch to `main`

Closes #207 

`eth-abi` previously only had typing for the `eth_abi/utils` folder. This PR adds typing for the `eth_abi/tools` folder, and adds the remaining files as a list to be excluded from `mypy` in `.pre-commit-config.yaml` As typing is added to individual files, those can be removed from the exclude list.

### Todo:

- [ ] replace `functools.lru_cache`  - `flake8-bugbear` rule B019
    - decided this deserves its own PR - Issue #218

- [x] Clean up commit history

#### Cute Animal Picture

image


eth_abi/exceptions.py
```diff
@@ -6,88 +6,67 @@ class EncodingError(Exception):
     Base exception for any error that occurs during encoding.
     """
 
-    pass
```
kclowes, line 9: Do we still need these `pass`es?

kclowes (approved):
Looks good! Just had a question about the removed `pass`es in the exeptions file.

eth_abi/exceptions.py
pacrob: They're removed by autoflake - https://github.com/PyCQA/autoflake - which says it removes "useless" `pass` statements.

A docstring renders a `pass` unnecessary: https://stackoverflow.com/questions/49754137/empty-class-with-comment-same-as-pass

I left it on default settings, but can put'm back if you want :smile:

kclowes: Oh cool. I'm all for default settings! 🚀

fjarri:
It seems that this PR deleted `py.typed` in the main `eth_abi` directory, thus breaking type checking in dependencies. Was that intentional?

pacrob:
@fjarri Sorry about that. I had moved the py.typed to the `tools` and `utils` folders because they were fully typed, while the `eth_abi/*.py` files only have minimal typing. I didn't think about the downstream breaking. I'll get that back in there.

fjarri:
No problem, just wanted to make sure it wasn't accidental. Technically the change appeared in v5.0 so API breakage is expected. I have no problem with doing some explicit `cast()` until `py.typed` can be returned.

PR #223, opened 2024-01-19
disallow payloads that decode to zero-sized types

pacrob:
### What was wrong?

Zero-sized types were disallowed in encoding in  #212. We should also disallow the decoding of payloads that decode to ZSTs.

### How was it fixed?

Find locations where dynamic types are decoded and check if the length is zero. If so, throw a `DecodingError`. Maybe define a new error type for this?

For reference, dynamic types are:

- bytes
- string
- dynamic-len arrays
- fixed-len arrays containing any dynamic type
- tuples containing any dynamic type

But we only need to be concerned with bytes, strings, and dynamic arrays, because they will be found within the other types as they are recursively decoded.

From https://docs.soliditylang.org/en/develop/abi-spec.html#formal-specification-of-the-encoding

Other things:
 - Removed warning about Solidity's V2 ABI encoder being experimental - it is now the default/only encoder: https://github.com/ethereum/solidity/pull/10336

### Todo:

- [ ] Clean up commit history

#### Cute Animal Picture

image


pacrob:
wrong approach here, starting over

PR #224, opened 2024-02-06
Add more typing

pacrob:
### What was wrong?

`eth-abi` has minimal typing for the `eth_abi/*.py` files. Looking to add to that and re-add the top-level `py.typed` file.

### How was it fixed?

Added some typing. Mainly changed the approach to gradual typing. Rather than ignore whole sets of files, run `mypy` on everything and ignore specific types of errors. This way the partial typing can still be considered.

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

image

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -48,4 +48,4 @@ repos:
     hooks:
     -   id: mypy
         args: [--follow-imports=silent, --disable-error-code=no-untyped-call]
-        exclude: tests\/|eth_abi\/(base\.py|registry\.py|grammar\.py|packed\.py|decoding\.py|encoding\.py|exceptions\.py|codec\.py|constants\.py)
+        exclude: tests/
--- a/eth_abi/decoding.py
+++ b/eth_abi/decoding.py
@@ -3,6 +3,7 @@ import decimal
 import io
 from typing import (
     Any,
+    Generator,
 )
 
 from eth_utils import (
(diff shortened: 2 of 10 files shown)


.pre-commit-config.yaml
```diff
@@ -48,4 +48,5 @@ repos:
     hooks:
     -   id: mypy
         args: [--follow-imports=silent, --disable-error-code=no-untyped-call]
-        exclude: tests\/|eth_abi\/(base\.py|registry\.py|grammar\.py|packed\.py|decoding\.py|encoding\.py|exceptions\.py|codec\.py|constants\.py)
+        # exclude: tests\/|eth_abi\/(base\.py|registry\.py|grammar\.py|packed\.py|decoding\.py|encoding\.py)
```
reedsa: Nit: Might not need this

reedsa (commented):
I'm in support of this! Is there any low hanging fruit that could be addressed for typing?

.pre-commit-config.yaml
pacrob: d'oh!

eth_abi/decoding.py
```diff
@@ -144,6 +145,7 @@ def decode(self, stream):
         start_pos = decode_uint_256(stream)
 
         stream.push_frame(start_pos)
+        assert self.tail_decoder is not None, "`tail_decoder is None"
```
kclowes: It's easy to ignore `assert`s in runtime code, so we tend to prefer to keep them only in tests. More info here: https://github.com/ethereum/snake-charmers-tactical-manual/blob/main/antipatterns.md#assertions-in-runtime-code

pacrob: update to use `AssertionError`

kclowes (approved):
Looks good to me! :rocket:

PR #226, opened 2024-02-07
Disallow invalid pointers in arrays and tuples

pacrob:
### What was wrong?

Incorrect values for in pointers can cause problems. If a pointer value is not large enough, i.e. it points to an area in the payload that is still within the pointers section, the encoding is malformed. In certain situations, ~infinite loops can occur.

### How was it fixed?
When decoding pointers, determine the location in the stream that divides pointers and values and make sure all pointers point past that location. Also check for pointers that point beyond the end of the payload.

Added some code comments to make it easier to remember how `HeadTailDecoder` works.

Added `pytest-timeout` to dependencies, as if the new tests are run without the added offset checking, they'll spin for a long time before failing.

### Todo:

- [x] Clean up commit history
- [x] Clear any breakpoints
- [x] clean up testing

#### Cute Animal Picture

image


eth_abi/decoding.py
```diff
@@ -131,6 +132,13 @@ def __call__(self, stream: ContextFramesBytesIO) -> Any:
 
 
 class HeadTailDecoder(BaseDecoder):
+    """
```
reedsa, line 135: 💯

eth_abi/decoding.py
```diff
@@ -172,8 +185,43 @@ def validate(self):
...
+            else:
+                # the next 32 bytes are not a pointer, so progress the stream per
+                # the decoder
+                decoder(stream)
+        # return the stream to its original location for actual decoding
+        stream.seek(current_location)
+
     @to_tuple  # type: ignore[misc] # untyped decorator
     def decode(self, stream: ContextFramesBytesIO) -> Generator[Any, None, None]:
+        self.validate_pointers(stream)
```
reedsa, line 224: I could use more context here. Could this be called in the loop below and maybe allow removal of the inner decoder loops inside `validate_pointers`? I'm also curious if the validation is necessary before decoding? Could validation just be part of the `decode` in `HeadTailDecoder`?

pacrob: There is no way to know how long the head section of a dynamic tuple will be until you have stepped through each decoder - if the decoder is for a dynamic type, it will be 32 bytes every time (because it's a pointer), but if it's for a non-dynamic array, there will be a single decoder for multiple chunks of 32 bytes.

I think it would be possible to take the logic from `validate_pointers` and put it in `decode` to eliminate the second loop through the decoders (where it actually checks the pointer values against the `end_of_offsets`). I like the current clarity and separation of concerns, but I can try if you like.

The validation needs to be in the tuple and array decoders, because only they have the context for how long they are. A `HeadTailDecoder` only has the info for a single dynamic value.

reedsa: I see now what the difference means, assuming there may never be more than a few decoders at a time I don't have any concerns.

eth_abi/decoding.py
```diff
@@ -172,8 +185,43 @@ def validate(self):
...
+        """
+        current_location = stream.tell()
+        len_of_head = sum(
+            decoder.array_size if hasattr(decoder, "array_size") else 1
+            for decoder in self.decoders
+        )
+        end_of_offsets = current_location + 32 * len_of_head
+        total_stream_length = len(stream.getbuffer())
+        for decoder in self.decoders:
+            if isinstance(decoder, HeadTailDecoder):
```
reedsa, line 200: Nit: It would be nice to share this logic across decoders, maybe this could become a utility function that could take the `stream` and an `array_size`, which could be called from here using `array_size=1`.

pacrob: Nit heard and politely declined. There is enough required difference in how tuples and arrays are checked that any logic extraction have a lot of `if tuple/elseif array`.  And I don't foresee any future datastructures being created that would make use of such shared base methods, thus accept code that is ~repeated twice.

kclowes (approved):
Looks good to me! Nice work tracking it down! 🐞 I like the comments you made in the decoder too. Very helpful.

PR #229, opened 2024-02-18
minor pep8 improvements

hkey0:
### How was it fixed?

Made minor pep8 improvements.


pacrob:
This is an internal script which doesn't need typing. We appreciate your effort, but won't be merging.

PR #241, opened 2025-01-02
Upgrade template

pacrob:
### What was wrong?

`py313` is in! And `bumpversion` is deprecated.

### How was it fixed?

Merged template, adding `py313` support, replacing `bumpversion` with `bump-my-version`, and running `mypy` locally instead of via the `pre-commit` mirror.

`hypothesis` still retained an upper pin that broke CI for `py312` and `313`, so added new pins `>=6.22.0,<6.108.7`. The new bottom pin is where all versions of python work and tested output is consistent with the upper pin. At the new upper pin, `<6.108.7`, they introduce new number formatting that broke some tests - https://hypothesis.readthedocs.io/en/latest/changes.html#v6-108-7 - They seem to break stuff in minors all the time, so probably good to keep the upper in place.

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

image


kclowes (approved):
🛥️ !

pacrob:
last force push changes making -> make in code of conduct

PR #243, opened 2025-02-14
docs: fix the installation package

matrixise:
### What was wrong?

The package's name to install is wrong in the documentation.


kclowes (approved):
Oops, thanks. Will add a newsfragment and merge.

PR #244, opened 2025-05-17
feat: cache result of ABIType.to_type_str

BobTheBuidler:
### What was wrong?
ABIType.to_type_str can be called multiple times on an instance, so its result should be cached for better performance

Related to Issue #
Closes #

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/eth_abi/grammar.py
+++ b/eth_abi/grammar.py
@@ -1,5 +1,6 @@
 import functools
 import re
+from typing import Optional
 
 import parsimonious
 from parsimonious import (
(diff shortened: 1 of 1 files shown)


eth_abi/grammar.py
```diff
@@ -291,22 +303,31 @@ def __init__(self, base, sub=None, arrlist=None, *, node=None):
...
+    
+            if isinstance(arrlist, tuple):
+                arrlist = "".join(repr(list(a)) for a in arrlist)
+            else:
+                arrlist = ""
+    
+            type_str = self._type_str = self.base + sub + arrlist
+        return type_str
 
     @property
```
BobTheBuidler, line 332: is there a reason we don't use cached_property here?

BobTheBuidler: It won't let me comment in the correct place, so I'll ask here. 

secondly, is there a reason we don't use cached_property on is_dynamic (or set it during init)?

BobTheBuidler:
I liked this PR a lot more before I found out about the pickling issue...

I'm not sure if it will actually be an improvement in its current form, some testing is needed. 

I might need to table this for now and make a PR in parsimonius repo to add a __hash__ method to Node objects

If nodes become hashable, this would again become a clear value-add

PR #251, opened 2025-12-10
feat(mypy): add some type hints in decoding.py

BobTheBuidler:
I noticed you're currently working on the repo and already had these ready from my work on [faster-eth-abi](https://github.com/BobTheBuidler/faster-eth-abi/), decided I'd contribute them back to the main repo so everybody can benefit!

I think I have most of the library properly typed. I could port over some other type hints too if I know it would be desired, but I don't want to spend time doing so if the PRs will sit there for 6 months. Lmk if this would be a welcome addition or not.

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture


BobTheBuidler:
I didn't add a release note for this PR, it seems unnecessary for this case but I can still add one if you prefer

PR #253, opened 2026-03-31
support hooks for offset detection

### LINKED ISSUE #252: Support hooks for offset detection
### What feature should we add?

When composing multiple on-chain protocols, there’s a requirement of calldata surgery to feed 0revious step results to the input of later steps. While the calldata is already encoded by the client, the on-chain logics usually just use offsets that’s provided by client to patch the calldata bytes.

Then the problem is how to generate the offsets, a typical approach is to encode a random value, and locate the random bytes in the output data to get offset, like this

```
v = random(32)
data = encode(sig, …, int.from_bytes(v), …)
idx = data.find(v)
```

The problems with this approach is:
1. The random value may collide with normal values
2. it don’t work with types with small bits, for example boolean.

A better approach is to support some hooks directly in the ABI library, then we can do something like this:

```
offsets = []
def needle(ctx):
  offsets.append((ctx.offset, ctx.size, ctx.type))
  return placeholder

encode(sig, …, needle, …, needle)
```

yihuang:
### What was wrong?

Closes #252

### Todo:

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/tests/core/packed/test_hooks.py
+++ b/tests/core/packed/test_hooks.py
@@ -75,7 +75,7 @@ def test_packed_hook_multiple_hooks():
         ["uint8", "uint8", "uint8"],
         [make_hook(1), make_hook(2), make_hook(3)],
     )
-    contexts == [
+    assert contexts == [
         EncodingContext(offset=0, size=1, encoder=registry_packed.get_encoder("uint8")),
         EncodingContext(offset=1, size=1, encoder=registry_packed.get_encoder("uint8")),
         EncodingContext(offset=2, size=1, encoder=registry_packed.get_encoder("uint8")),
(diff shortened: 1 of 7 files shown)


tests/core/packed/test_hooks.py
```diff
@@ -0,0 +1,365 @@
...
+            contexts.append(ctx)
+            return retval
+
+        return hook
+
+    resolved = encode_packed_with_hooks(
+        ["uint8", "uint8", "uint8"],
+        [make_hook(1), make_hook(2), make_hook(3)],
+    )
+    contexts == [
```
yihuang:
```suggestion
    assert contexts == [
```

PR #257, opened 2026-05-26
fix(ci): route Windows wheel jobs to tox env

BobTheBuidler:
- Route `windows-latest` wheel matrix jobs to the existing `windows-wheel` tox env.
- Keep non-Windows and core jobs on `py3<minor>-<jobtype>`.
- Avoid running the Unix `py3xx-wheel` tox env on Windows.


BobTheBuidler:
dude shit my bad again

